PR #1, opened 2018-01-25
Add azure templates project to solution

EwanNoble:


EwanNoble:
@chelnak  These have now been removed.

PR #4, opened 2018-02-12
ER-240 vacancy creation

chrisjensenuk:
- added Controller, ViewModel and View stubs
- globally added AutoValidateAntiforgeryTokenAttribute. Use [IgnoreAntiforgeryToken] to ignore on an action. 
- added attribute validation 
- Using IActionResult rather than ActionResult - no async as yet
- ValidationSummary is always renedered even if valid - needs to be hidden in css - .validation-summary-valid
- Added pattern - ViewModels and EditModels - see RoleDescription for example

There was some discussion about the use of tag helpers for routing being brittle. Maybe route names would be better? We change change this later.


leewadhams:
Looks good.
I've created an ErrorController and got rid of the models folder.

PR #14, opened 2018-02-15
ER-240 orchestrators and mongo repository

chrisjensenuk:
- added an orchestrator for creating a new vacancy
- Using IOptions<> for mongo configuration injection
- created a StubRepository which is used if you don't have mongo configuration information
- created MongoDbVacancyRepository to persist a new vacancy to  mongo
- Using Mongo mapping configuration instead of attributes so can keep the Vacancy entity clean of Mongo stuff.

I'll do a demo tomorrow to walk you all though this

## Modifications since demo Friday 16/02
- Mongo config - now using connection string
- Orchestrator - removed interface
- Handlers must not return responses - removed response
- Guid Id now being generated by our application and not Mongo
- Mongo entity mapping - now using convention rather than mapping for camelCasing
- Added IoC registry to Recruit.Storage.Client

## Modifications since demo Monday 19/02
- Now using a single upsert handler that inserts/updates via patches
- Added an orchestrator for each controller


src/Employer.Web/Controllers/NewVacancyController.cs
```diff
@@ -1,24 +1,24 @@
...
         {
-            _messaging = messaging;
+            _orchestrator = orchestrator;
         }
         
         [HttpGet, Route("accounts/{employerAccountId}/new-vacancy", Name = RouteNames.NewVacancy_Index_Get)]
         public IActionResult Index()
         {
-            var vm = new IndexViewModel();
+            var vm = _orchestrator.GetIndexViewModel();
```
leewadhams: My preference would be to make the naming of the orchestrator methods reflect the use-case. Am happy to go with the majority though.

src/Employer.Web/Controllers/NewVacancyController.cs
```diff
@@ -1,24 +1,24 @@
...
+        private readonly INewVacancyOrchestrator _orchestrator;
 
-        public NewVacancyController(IMessaging messaging)
+        public NewVacancyController(INewVacancyOrchestrator orchestrator)
         {
-            _messaging = messaging;
+            _orchestrator = orchestrator;
         }
         
         [HttpGet, Route("accounts/{employerAccountId}/new-vacancy", Name = RouteNames.NewVacancy_Index_Get)]
```
leewadhams: Another one for discussion but I'd vote for putting the common part of the route at the class level.

src/Employer.Web/Orchestrators/INewVacancyOrchestrator.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Esfa.Recruit.Employer.Web.ViewModels.NewVacancy;
```
leewadhams: There probably is an argument for not having the interface for an orchestrator as they are tightly coupled to the controller, i.e. not very likely we'll want to use them anywhere else.

Although may be useful if unit testing at the controller level for any reason.

src/Recruit.Storage.Client/Core/Entities/MongoVacancy.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Esfa.Recruit.Storage.Client.Core.Entities
+{
+    public class MongoVacancy
+    {
+        [BsonId]
+        public Guid Id { get; private set; }
+
+        [BsonElement("vrn")]
+        public int VRN { get; set; }
```
leewadhams: I can't believe you listened to @mgwilliam 😄

src/Recruit.Storage.Client/Core/Mongo/MongoDbCollectionBase.cs
```diff
@@ -0,0 +1,46 @@
...
+        private const string MongoCredentialMechanism = "SCRAM-SHA-1";
+
+        public MongoDbCollectionBase(string dbName, string collectionName, IOptions<MongoDbConnectionDetails> details)
+        {
+            _dbName = dbName;
+            _collectionName = collectionName;
+            _details = details.Value;
+        }
+
+        protected IMongoCollection<T> GetCollection<T>()
```
leewadhams: I'm wondering if it would make things simpler if we create the client from a mongo connection string:

```
string connectionString = 
  @"mongodb://<user>:<password>@<host>:<port>/?ssl=true&replicaSet=globaldb";

MongoClientSettings settings = MongoClientSettings.FromUrl(
  new MongoUrl(connectionString)
);
```

src/Recruit.Storage.Client/Core/Mongo/MongoDbConventions.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Esfa.Recruit.Storage.Client.Core.Mongo
+{
+    public static class MongoDbConventions
+    {
+        public static void RegisterMongoConventions()
+        {
+            var pack = new ConventionPack();
+            pack.Add(new CamelCaseElementNameConvention());
+            ConventionRegistry.Register("camelCase", pack, t => true);
```
sachpatel: Does this mean the name of the field in Mongo will always be the same as in the codebase apart from different case? As in we can;t change it?

src/Recruit.Storage.Client/Core/Entities/Vacancy.cs
```diff
@@ -1,7 +1,13 @@
+﻿using System;
+
 namespace Esfa.Recruit.Storage.Client.Core.Entities
 {
     public class Vacancy
     {
+        public Guid Id { get; set; }
         
+        public int VRN { get; set; }
```
sachpatel: I am not against having a shorter name as discussed yesterday. Since we will have a collection for `Vacancy` I can understand ommiting the word Vacancy from the field but maybe we call it `Reference` instead, that way from looking at the database it is easier to interpret what that field is for without knowing about the codebase. `vrn` is a bit ambiguous but I understand your keystrokes reasoning.

src/Recruit.Storage.Client/Core/Ioc/Registry.cs
```diff
@@ -0,0 +1,40 @@
...
+using Esfa.Recruit.Storage.Client.Core.Services;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using MediatR;
+
+namespace Esfa.Recruit.Storage.Client.Core.Ioc
+{
+    public static class Registry
+    {
+        public static void RegisterRecruitStorageClient(this IServiceCollection services, IConfiguration configuration)
```
leewadhams: Small point. Should we follow the convention of Add***() for adding services?

src/Recruit.Storage.Client/Core/Mongo/MongoDbConventions.cs
chrisjensenuk: Yep. At the demo we thought using conventions over a mapping for each property would be safer.  We can always add explicit mappings if we need them

src/Employer.Web/Configuration/IoC.cs
```diff
@@ -1,54 +1,28 @@
...
-using MongoDB.Bson.Serialization;
 
 namespace Esfa.Recruit.Employer.Web.Configuration
 {
     public static class IoC
     {
         public static void AddIoC(this IServiceCollection services, IConfiguration configuration)
         {
+            //Register Recruit.Storage.Client IoC
+            services.RegisterRecruitStorageClient(configuration);
```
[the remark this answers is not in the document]
chrisjensenuk: This isn't adding a service as such. Its calling the register which in turn adds services so thought I'd use a different name when adding a register.

src/Employer.Web/Configuration/IoC.cs
```diff
@@ -12,14 +12,29 @@ public static class IoC
     {
         public static void AddIoC(this IServiceCollection services, IConfiguration configuration)
         {
+            //Recruit.Storage.Client IoC
```
sachpatel: 💣 💥

src/Employer.Web/Controllers/NewVacancyController.cs
```diff
@@ -1,46 +1,40 @@
...
 using Esfa.Recruit.Employer.Web.Configuration.Routes;
-using Esfa.Recruit.Storage.Client.Core.Messaging;
 using Esfa.Recruit.Storage.Client.Core.Commands;
+using Esfa.Recruit.Employer.Web.Orchestrators;
+using System.Threading.Tasks;
 
 namespace Esfa.Recruit.Employer.Web.Controllers
 {    
+    [Route("accounts/{employerAccountId}")]
     public class NewVacancyController : Controller
```
sachpatel: Could we call this `CreateVacancyController`. Me being picky but something about the `New` word makes me shudder 😕

src/Recruit.Storage.Client/Core/Mongo/MongoDbConventions.cs
sachpatel: 👍

src/Recruit.Storage.Client/Core/Repositories/StubVacancyRepository.cs
```diff
@@ -0,0 +1,35 @@
...
+            {
+                _vacancy = new Vacancy
+                {
+                    Id = vacancyId
+                };
+            }
+
+            foreach(var property in patch.GetType().GetProperties())
+            {
+                typeof(Vacancy).GetProperty(property.Name).SetValue(_vacancy, property.GetValue(patch));
```
sachpatel: 👨‍🔬

src/Recruit.Storage.Client/Core/Repositories/MongoDbVacancyRepository.cs
```diff
@@ -0,0 +1,46 @@
...
+using System;
+using System.Threading.Tasks;
+
+namespace Esfa.Recruit.Storage.Client.Core.Repositories
+{
+    public class MongoDbVacancyRepository : MongoDbCollectionBase, ICommandVacancyRepository, IQueryVacancyRepository
+    {
+
+        private const string Database = "Recruit";
+        private const string Collection = "Vacancy";
```
sachpatel: I think it is common to use the plural for collection names and lowercase. http://arkusnexus.com/2016/09/12/coding-guidelines-mongodb/?nabe=5802299602763776:0

src/Employer.Web/Configuration/IoC.cs
```diff
@@ -12,14 +12,29 @@ public static class IoC
...
             services.Configure<ExternalLinksConfiguration>(configuration.GetSection("ExternalLinks"));
+            
             services.AddSingleton<IHttpContextAccessor, HttpContextAccessor>(); // Used by NLog to log out traceidentifier value.
 
             services.AddTransient<IGetAssociatedEmployerAccountsService, GetAssociatedEmployerAccountsService>();
 
-            //Mediatr
-            services.AddMediatR(typeof(CreateVacancyCommandHandler).Assembly);
-            services.AddTransient<IMessaging, MediatrMessaging>();
+            //Orchestrators
```
sachpatel: I'm going to be shot down 🔫 as usual but `#region` #justsayin 😋 

Because it is static we can't even move it to a partial class 😞

chrisjensenuk: At some point we may need to do this.  Its not too long at the mo so I'm leaving it for now.  Maybe we could group similar setups into their own .Addxxx extensions. E.g. services.AddOrchestrators()

src/Recruit.Storage.Client/Core/Repositories/MongoDbVacancyRepository.cs
chrisjensenuk: 👍

src/Employer.Web/Controllers/NewVacancyController.cs
chrisjensenuk: Will do.

src/Employer.Web/Configuration/IoC.cs
chrisjensenuk: 👍

PR #19, opened 2018-02-21
Feature/ER-219 Dashboard list vacancies

sachpatel:


mgwilliam:
@sachpatel the dashboard view model should be populated with data retrieved from the query store not the vacancy repository. are you planning to do this in a separate PR or ticket?

src/Employer.Web/Configuration/ConfigurationExtensions.cs
```diff
@@ -98,7 +100,7 @@ public static void AddAuthenticationService(this IServiceCollection services, Au
         private static Task PopulateAccountsClaim(Microsoft.AspNetCore.Authentication.OpenIdConnect.TokenValidatedContext ctx, IGetAssociatedEmployerAccountsService accountsSvc)
         {
             var userId = ctx.Principal.Claims.First(c => c.Type.Equals(EmployerRecruitClaims.IdamsUserIdClaimTypeIdentifier)).Value;
-            var accounts = accountsSvc.GetAssociatedAccounts(userId);
+            var accounts = accountsSvc.GetAssociatedAccounts(userId).Result;
```
chrisjensenuk: I think we may be able to async the calling function so we don't have to wait for the result. options.Events.OnTokenValidated = async (ctx) => PopulateAccountsClaim(ctx, accountsSvc);

src/Employer.Web/Services/GetAssociatedEmployerAccountsService.cs
```diff
@@ -1,10 +1,49 @@
...
+        private readonly IAccountApiClient _accountApiClient;
+
+        public GetAssociatedEmployerAccountsService(ILogger<GetAssociatedEmployerAccountsService> logger, IAccountApiClient accountApiClient)
         {
-            return new string[] { "abc", "xyz" };
+            _logger = logger;
+            _accountApiClient = accountApiClient;
+        }
+
+        public async Task<string[]> GetAssociatedAccounts(string userId)
```
chrisjensenuk: I've appended 'Async' to my async functions so GetAssociatedAccountsAsync

src/Employer.Web/Configuration/IoC.cs
```diff
@@ -16,24 +18,45 @@ public static void AddIoC(this IServiceCollection services, IConfiguration confi
...
+
+        private static void RegisterAccountApiClientDeps(IServiceCollection services)
+        {
+            services.AddSingleton<IAccountApiConfiguration>(kernal => kernal.GetService<IOptions<AccountApiConfiguration>>().Value);
+            services.AddTransient<IAccountApiClient, AccountApiClient>();
+        }
+
+        private static void RegisterOrchestratorDeps(IServiceCollection services)
+        {
+            services.AddTransient<ApplicationProcessOrchestrator>();
```
chrisjensenuk: Ta!

src/Employer.Web/Configuration/IoC.cs
```diff
@@ -16,24 +18,45 @@ public static void AddIoC(this IServiceCollection services, IConfiguration confi
             //Configuration
             services.Configure<ExternalLinksConfiguration>(configuration.GetSection("ExternalLinks"));
             services.Configure<AuthenticationConfiguration>(configuration.GetSection("Authentication"));
+            services.Configure<AccountApiConfiguration>(configuration.GetSection("AccountApiConfiguration"));
+            
+            RegisterAccountApiClientDeps(services);
```
chrisjensenuk: I'm liking extensions for stuff like this so services.AddAccountApiClientDeps();

src/Employer.Web/Services/GetAssociatedEmployerAccountsService.cs
```diff
@@ -1,10 +1,49 @@
...
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Esfa.Recruit.Employer.Web.Services
 {
     public class GetAssociatedEmployerAccountsService : IGetAssociatedEmployerAccountsService
     {
-        public string[] GetAssociatedAccounts(string userId)
+        private readonly ILogger<GetAssociatedEmployerAccountsService> _logger;
```
chrisjensenuk: Do we need the 'Get'. Would just AssociatedEmployerAccountsService suffice?

src/Employer.Web/Configuration/IoC.cs
```diff
@@ -16,24 +18,45 @@ public static void AddIoC(this IServiceCollection services, IConfiguration confi
...
 
             services.AddSingleton<IHttpContextAccessor, HttpContextAccessor>(); // Used by NLog to log out traceidentifier value.
 
-            services.AddTransient<IGetAssociatedEmployerAccountsService, GetAssociatedEmployerAccountsService>();
+            if (configuration.GetValue<bool>("UseStubs"))
```
chrisjensenuk: Do we need to add this configuration option?  For the Mongo repo I've just checked to see if I have Mongo configuration and if I haven't then I just use the stub.  Can/should we do something similar here?

src/Employer.Web/Configuration/IoC.cs
sachpatel: I probably wouldn't for the moment as I think we are changing stuff or well I will be. But have no issue with sticking it in extension methods later on. I can understand where you have for the Recruit client thingy.

src/Employer.Web/Configuration/IoC.cs
sachpatel: I was trying it but felt it wasn't explicit enough for me. This `UseStubs` approach isn't something I am enforcing but found while working disconnected was going to suit me better for the moment.

src/Employer.Web/Controllers/CreateVacancyController.cs
```diff
@@ -25,14 +24,14 @@ public IActionResult Index()
         }
 
         [HttpPost("new-vacancy", Name = RouteNames.NewVacancy_Index_Post)]
-        public async Task<IActionResult> Index(IndexViewModel vm)
+        public async Task<IActionResult> Index(string employerAccountId, IndexViewModel vm)
```
chrisjensenuk: For the vacancyId work I've done I've just added it as a property to the viewmodel and the Model binder automatically populates it from the route.  I like only having one parameter being posted to an action - less ambiguous.

src/Employer.Web/Orchestrators/DashboardOrchestrator.cs
```diff
@@ -0,0 +1,38 @@
...
+        private readonly IGetAssociatedEmployerAccountsService _getAccountService;
+
+        public DashboardOrchestrator(IMessaging messaging, IGetAssociatedEmployerAccountsService getAccountsService, IQueryVacancyRepository queryRepository)
+        {
+            _messaging = messaging;
+            _getAccountService = getAccountsService;
+            _queryRepository = queryRepository;
+        }
+
+        public async Task<DashboardViewModel> GetDashboardViewModelAsync(string employerAccountId)
```
chrisjensenuk: Obvs this bit will change when QueryStore is introduced.  I'm expecting the returned document for the QS can pretty much be one-to-one mapped to the view model.

src/Employer.Web/Services/StubGetAssociatedEmployerAccountsService.cs
```diff
@@ -0,0 +1,21 @@
...
+        public Task<string[]> GetAssociatedAccounts(string userId)
+        {
+            return Task.FromResult(new string[] { "abc", "xyz", "MYJR4X", "MB6YDY" });
+        }
+
+        public Task<AccountDetailViewModel> GetEmployerAccountAsync(string employerAccountId)
+        {
+            return Task.FromResult(new AccountDetailViewModel
+            {
+                DasAccountName = "Ozzy Scott"
```
chrisjensenuk: Aussie?

src/Employer.Web/Controllers/CreateVacancyController.cs
sachpatel: Will have a think about this but sounds good.

src/Employer.Web/Orchestrators/DashboardOrchestrator.cs
sachpatel: Yep, I would rather get this work in and then do the QS stuff as a separate piece.

src/Employer.Web/Services/GetAssociatedEmployerAccountsService.cs
sachpatel: This was because of the stub intention. Renamed it to `EmployerAccountService`

src/Employer.Web/Services/GetAssociatedEmployerAccountsService.cs
sachpatel: 👍

PR #21, opened 2018-02-23
Update azure template with new config value for mongo

leewadhams:
Ewan, can you have a quick check that the way i've added the new connectionString is correct.

Are you happy for us to add new config values into the template going forward?


EwanNoble:
@leewadhams Hi Lee, yes that's fine, I will still need to set the parameters though so if you could continue to include me on PRs that would be great. In this particular case I can probably do something clever to avoid a parameter here, will have an implementation shortly.

PR #22, opened 2018-02-23
ER-220 submit navigation

chrisjensenuk:
Changes the status of a Vacancy to submitted.  A submitted vacancy cannot be re-submitted.

- I've added VacancyExtensions.cs to keep Vacancy.cs skinny. Is this how we want to add helper functionality to the Vacancy?
- For Created and Submitted dates do we want to use DateTime.UtcNow or just DateTime.Now?
- I've added a convention so Enums are persisted to Mongo as strings


src/Recruit.Vacancies.Client/Domain/Entities/Vacancy.cs
```diff
@@ -8,7 +9,12 @@ public class Vacancy
         
         public string Title { get; set; }
         
-        public DateTime? AuditVacancyCreated { get; set; }
         public global::System.String EmployerAccountId { get; internal set; }
```
sachpatel: 😕

src/Recruit.Vacancies.Client/Domain/Entities/VacancyExtensions.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Esfa.Recruit.Storage.Client.Domain.Enum;
+
+namespace Esfa.Recruit.Storage.Client.Domain.Entities
+{ 
+    public static class VacancyExtensions
+    {
+        public static bool IsSubmittable(this Vacancy vacancy)
```
sachpatel: I think having this as a read-only property on the model would be fine e.g. `public bool IsSubmittale => Status == VacancyStatus.Draft;`

Its not complex logic or behavior you are introducing so I think it is fine to keep in Entity.Also you are only inspecting the state it already knows about, not the state of another object.

src/Recruit.Vacancies.Client/Domain/Enums/VacancyStatus.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Esfa.Recruit.Storage.Client.Domain.Enum
```
sachpatel: I don't think we need a separate namespace for enum.

src/Employer.Web/wwwroot/css/application.css
```diff
@@ -496,4 +496,9 @@ li.menu-link.sub-menu-open ul li a {
 
 .actions .action {
   margin-bottom: 20px;
+}
+
+/***  Inline form ***/
+.form-inline{
```
sachpatel: 🎨 - Is this because of that inline style CSP error?

src/Employer.Web/ViewModels/Preview/IndexViewModel.cs
```diff
@@ -8,5 +8,7 @@ namespace Esfa.Recruit.Employer.Web.ViewModels.Preview
     public class IndexViewModel
     {
         public string Title { get; set; }
+
+        public bool IsSubmittable { get; set; }
```
sachpatel: Probably call this `CanSubmit` maybe, save some keystrokes 😉

src/Recruit.Vacancies.Client/Domain/Entities/Vacancy.cs
chrisjensenuk: What happened there?!? Will fix

src/Recruit.Vacancies.Client/Domain/Entities/VacancyExtensions.cs
chrisjensenuk: Sounds like a plan. Will change

src/Recruit.Vacancies.Client/Domain/Enums/VacancyStatus.cs
chrisjensenuk: Agree. Will fix

src/Employer.Web/wwwroot/css/application.css
chrisjensenuk: Nope this is because the user-agent by default renders <form> as a block element.  I want to hide the whole form if a vacancy is already submitted (so the form and submit button are never rendered) but by doing this the styling was screwing up.

src/Employer.Web/ViewModels/Preview/IndexViewModel.cs
chrisjensenuk: I agonised over this. I originally called it CanSubmit but tried squeeze an 'Is' on to the start of it.  I'm happier with CanSubmit if you are - thought you were more of an 'Is' person. Will change

src/Employer.Web/Controllers/PreviewController.cs
```diff
@@ -24,10 +24,22 @@ public PreviewController(PreviewOrchestrator orchestrator)
             return View(vm);
         }
 
-        [HttpPost("vacancy-preview", Name = RouteNames.Preview_Index_Post)]
-        public IActionResult Index(IndexViewModel vm)
+        [HttpPost("vacancy-submit", Name = RouteNames.Preview_Submit_Post)]
+        public async Task<IActionResult> Submit(SubmitEditModel m)
         {
```
sachpatel: If this action is Submit(Not preview), why do we need to post to it when the vacancyId is in the url. I think maybe a get would be suitable. That way on the preview page you can just choose to render a link for the button instead of the form. You already have a conditional around whether to show it.

src/Employer.Web/ViewModels/Preview/IndexViewModel.cs
sachpatel: Thanks for considering me a person. Not too bothered but thought about your keystrokes reasoning.

src/Employer.Web/Controllers/PreviewController.cs
sachpatel: ok

PR #30, opened 2018-02-26
ER-250 Delete Draft Vacancy

chrisjensenuk:
- Added handler to manage the deletion of draft vacancies
- Added new 'are you sure?' page which is displayed when clicking the 'delete' link

### DIFF
--- a/src/Employer.Web/Views/DeleteVacancy/Delete.cshtml
+++ b/src/Employer.Web/Views/DeleteVacancy/Delete.cshtml
@@ -7,23 +7,26 @@
         <h1 class="heading-large">@Model.Title</h1>
         <h1 class="heading-large">Delete vacancy</h1>
         
-        <h2 class="heading-medium">Are you sure?</h2>
-
         <form asp-route="@RouteNames.DeleteVacancy_Delete_Post">
             <div class="form-group">
-                <div class="multiple-choice">
-                    <input asp-for="@Model.ConfirmDeletion" type="radio" value="true" id="deleteYes" />
-                    <label class="block-label selection-button-radio" for="deleteYes">Yes, delete now</label>
-                </div>
-                <div class="multiple-choice">
-                    <input asp-for="@Model.ConfirmDeletion" type="radio" value="false" id="deleteNo" />
-                    <label class="block-label selection-button-radio" for="deleteNo">No, don't delete</label>
-                </div>
+                <fieldset>
+                    <legend class="heading-medium">Are you sure?</legend>
+
+                    <div class="multiple-choice">
+                        <input asp-for="@Model.ConfirmDeletion" type="radio" value="true" id="delete-yes" />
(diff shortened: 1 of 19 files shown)


src/Recruit.Vacancies.Client/Domain/Entities/Vacancy.cs
```diff
@@ -17,6 +17,17 @@ public class Vacancy
...
         public DateTime? SubmittedDate { get; set; }
 
-        public bool CanSubmit => Status == VacancyStatus.Draft;
+        public bool IsDeleted { get; set; }
+        public DateTime? DeletedDate { get; set; }
+
+        /// <summary>
+        /// We can only submit draft vacancies that have not been deleted
+        /// </summary>
+        public bool CanSubmit => Status == VacancyStatus.Draft && IsDeleted == false;
```
sachpatel: Do these field get persisted to the Vacancy in Mongo/Cosmos as well?

src/Recruit.Vacancies.Client/Domain/Entities/Vacancy.cs
```diff
@@ -17,6 +17,17 @@ public class Vacancy
 
         public DateTime? SubmittedDate { get; set; }
 
-        public bool CanSubmit => Status == VacancyStatus.Draft;
+        public bool IsDeleted { get; set; }
```
sachpatel: Shouldn't IsDeleted be derived from Status?

src/Employer.Web/Views/DeleteVacancy/Delete.cshtml
```diff
@@ -0,0 +1,32 @@
...
+        <h1 class="heading-large">@Model.Title</h1>
+        <h1 class="heading-large">Delete vacancy</h1>
+        
+        <form asp-route="@RouteNames.DeleteVacancy_Delete_Post">
+            <div class="form-group">
+                <fieldset>
+                    <legend class="heading-medium">Are you sure?</legend>
+
+                    <div class="multiple-choice">
+                        <input asp-for="@Model.ConfirmDeletion" type="radio" value="true" id="delete-yes" />
```
sachpatel: Do we need to create our own `id` attribute, is it for styling? I thought the `asp-for` would generate it for us.

src/Recruit.Vacancies.Client/Domain/Entities/Vacancy.cs
chrisjensenuk: Nope. Just checked

src/Recruit.Vacancies.Client/Domain/Entities/Vacancy.cs
chrisjensenuk: Had a chat to Mark about this this morning. We want to know what status the vacancy was in before it was deleted so using a separate bool field for deletions

PR #31, opened 2018-02-27
Feature/ER-219 Dashboard from query store

sachpatel:
Have used a separate mongo db for storing dashboard data in a collection. The collection is called dashboard-views. I know we may change this and use a single collection for all views but so far we only have one purpose.

I have used the EmployerAccountId as the identifier on the dashboard stored. Was going to use a key like `{EmployerAccountId}_Dashboard` but again thought it isn't needed yet and so have kept it simple.

Have introduced mapping classes to keep it separate but am instantiating concrete versions in consuming classes as we don;t have a need for mocking or separating these yet.


src/Employer.Web/Orchestrators/DashboardOrchestrator.cs
```diff
@@ -1,33 +1,29 @@
...
-            
-            var vm = new DashboardViewModel
-            {
-                EmployerName = account.Result.DasAccountName,
-                Vacancies = vacancies.Result.OrderByDescending(v => v.CreatedDate).ToList()
-            };
+            var dashboard = _vacancyClient.GetDashboardAsync(employerAccountId);
+            await Task.WhenAll(account, dashboard);
+
+            var vm = DashboardMapper.MapFromDashboard(dashboard.Result, account.Result);
```
chrisjensenuk: I would have been tempted to turn this into an extension method off Dashboard.

src/Recruit.Vacancies.Client/Application/Handlers/UpdateDashboardCommandHandler.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            var key = string.Format(QueryViewKeys.DashboardViewPrefix, message.EmployerAccountId);
+            var employerVacancies = await _repository.GetVacanciesByEmployerAccountAsync(message.EmployerAccountId);
+            var vacancySummaries = employerVacancies.Select(v => VacancySummaryMapper.MapFromVacancy(v));
+            
+            var newDashboard = CreateNewDashboard(key, vacancySummaries);
+            await _writer.UpdateDashboardAsync(key, newDashboard);
+        }
+
+        private Dashboard CreateNewDashboard(string key, IEnumerable<VacancySummary> summaries)
```
chrisjensenuk: CreateDashboard maybe?

src/Recruit.Vacancies.Client/Application/Handlers/UpdateDashboardCommandHandler.cs
```diff
@@ -0,0 +1,47 @@
...
+            _repository = repository;
+        }
+
+        public async Task Handle(UpdateDashboardCommand message, CancellationToken cancellationToken)
+        {
+            var key = string.Format(QueryViewKeys.DashboardViewPrefix, message.EmployerAccountId);
+            var employerVacancies = await _repository.GetVacanciesByEmployerAccountAsync(message.EmployerAccountId);
+            var vacancySummaries = employerVacancies.Select(v => VacancySummaryMapper.MapFromVacancy(v));
+            
+            var newDashboard = CreateNewDashboard(key, vacancySummaries);
```
chrisjensenuk: Do we need the 'new'? var dashboard =

src/Recruit.Vacancies.Client/Infrastructure/Client/VacancyClient.cs
```diff
@@ -12,24 +14,25 @@ namespace Esfa.Recruit.Vacancies.Client.Infrastructure.Client
...
+        public async Task UpdateVacancyAsync(Vacancy vacancy)
         {
             var command = new UpdateVacancyCommand
             {
                 Vacancy = vacancy
             };
 
-            return _messaging.SendCommandAsync(command);
+            await _messaging.SendCommandAsync(command);
+            await UpdateDashboardAsync(vacancy.EmployerAccountId);
```
chrisjensenuk: Should we update the dashboard every time we update a vacancy?  Maybe we should ass in a bool so the orchestrators can specify that they have changed a field that affects the dashboard. Dunno.  One for Mark.

src/Recruit.Vacancies.Client/Infrastructure/Client/VacancyClient.cs
```diff
@@ -98,8 +102,25 @@ public Task UpdateVacancyAsync(Vacancy vacancy)
...
 
             await _messaging.SendCommandAsync(command);
+            await UpdateDashboardAsync(vacancy.EmployerAccountId);
 
             return true;
         }
+        
+        public async Task<Dashboard> GetDashboardAsync(string employerAccountId)
+        {
+            var key = string.Format(QueryViewKeys.DashboardViewPrefix, employerAccountId);
```
chrisjensenuk: Do we need a prefix?

src/Recruit.Vacancies.Client/Infrastructure/QueryStore/QueryStore.cs
```diff
@@ -1,32 +1,46 @@
...
             : base(Database, Collection, details)
         {
         }
 
-        public async Task<IEnumerable<Vacancy>> GetVacanciesAsync(string employerAccountId)
+        public async Task<Dashboard> GetDashboardAsync(string key)
         {
-            var filter = Builders<Vacancy>.Filter.Eq(v => v.EmployerAccountId, employerAccountId);
-            filter = filter & Builders<Vacancy>.Filter.Eq(v => v.IsDeleted, false);
+            var filter = Builders<Dashboard>.Filter.Eq(d => d.ViewKey, key);
```
chrisjensenuk: As its mongo shouldn't we be targeting the 'id' field?

src/Recruit.Vacancies.Client/Infrastructure/QueryStore/QueryStore.cs
```diff
@@ -1,32 +1,46 @@
...
 
-            var collection = GetCollection<Vacancy>();
+            var collection = GetCollection<Dashboard>();
             var result = await collection.FindAsync(filter);
 
-            return result.ToEnumerable();
+            return result?.FirstOrDefault();
+        }
+
+        public async Task UpdateDashboardAsync(string key, Dashboard dashboard)
```
chrisjensenuk: Can we use 'key' as our 'id' in Mongo? Can we use a custom ObjectId so we can use EmployerAccountId as the id?

mgwilliam: you can just add the bsonid attribute or configure in the bson class mapper if you want a custom id OR just call the field "id"

src/Recruit.Vacancies.Client/Infrastructure/Client/VacancyClient.cs
sachpatel: I am not sure, it's difficult because this could easily be forgotten about in other places that might need it. I don't really like having to do these two things but that is what has been suggested.

src/Recruit.Vacancies.Client/Infrastructure/QueryStore/QueryStore.cs
sachpatel: Yeah I had issues with this as Mongo requires the Id to be untouched but at the same time it doesn't make identifying the item in Mongo easier.

src/Recruit.Vacancies.Client/Infrastructure/Client/VacancyClient.cs
sachpatel: Again, I am not sure I did this because otherwise we may have other views for an employer that isn't data to do with a dashboard and this makes it seem a bit more key/valuey even though it isn;t plain key/value.

src/Recruit.Vacancies.Client/Infrastructure/QueryStore/QueryStore.cs
chrisjensenuk: If we make the id the EmployerAccountId its not going to change so will remain untouched

src/Recruit.Vacancies.Client/Infrastructure/Client/VacancyClient.cs
mgwilliam: if we did add a flag then i'd say it should update by default

sachpatel: so `bool forceUpdate = true` as the last parameter?

chrisjensenuk: updateQueryStores?

src/Recruit.Vacancies.Client/Infrastructure/Client/VacancyClient.cs
mgwilliam: the idea of the composite key (type + subject) is that it avoids us having to create separate collections for every different type of projection as there will be lots of them in the app further down the line. it's also inherently unique so could be used as the document id as chris suggested elsewhere in this pr.

i'd also originally thought we should add a "type" field in there so that we could (for example) retrieve all the vacancy projects. i suspect we will end up needing this but isn't strictly needed right now.

sachpatel:
Hi I have updated the PR taking into account comments made. Regarding not calling the accounts API for every dashboard page load it is now taken from the claim which is populated when the user is first authenticated (per session). 

The claim has changed structure and is now a json object representing a dictionary with a complex type for `EmployerIdentifier` details. This was done rather than having more claims because we are unsure of which account the user is acting under when they are authenticating from Idams. We can change this so each employer account and name is stored in individual claims so we do not need to serialize/deserialize an object. I have just done this for the moment.

Alternative claim structure is to store the value as a plain string e.g. '`AccountId`:`EmployerName`'

PR #36, opened 2018-03-06
Fix for grabbing App Insights InstrumentationKey from config

sachpatel:
I had an issue running the web role because of a null exception being logged when trying to use App Insights. I found that the key specified needed to be changed. Able to run the TrainingType job successfully now.


sachpatel:
Not needed.

PR #38, opened 2018-03-06
Add collections to cosmos config file

leewadhams:
Although the database and collections are automatically created if they don't exist we need to be able to pre create them in the release to set things like the thoughput amount (RUs) and partition key.

Therefore we need to add any new collections into the config file.


leewadhams:
Put change into folder restructure pull request #39

PR #40, opened 2018-03-08
:speaker: - Adding handling for application exceptions

sachpatel:
This is a bit of a crude method.

I have looked at using IExceptionFilter's or our own custom Exception middleware for handling of global exceptions we raise. Both options were not quite right or appropriate.

I have only put the exception raising in one place at the moment which is when attempting to go to the `sections` page for a vacancy that has `submitted` status.

I have placed the example `ConcurrencyException` in the web project as I was not sure where to put it if at all in the VacancyClient project application or domain layer folders.

Because we want to surface an error such as the one for trying to edit a submitted vacancy on the dashboard, I have made the dashboard a Shared view that can be called to display from any controller.

To see in action, in the `Startup.Configure.cs` file, replace `app.UseDeveloperExceptionPage();` with `app.UseExceptionHandler("/error/handle");` and then click onto the title of a vacancy that has a `Submitted` status from the dashboard.


src/Employer/Employer.Web/Orchestrators/SectionsOrchestrator.cs
```diff
@@ -18,6 +20,11 @@ public SectionsOrchestrator(IVacancyClient client)
         {
             var vacancy = await _client.GetVacancyForEditAsync(vacancyId);
 
+            if (vacancy.Status == VacancyStatus.Submitted)
+            {
+                throw new ConcurrencyException($"The vacancy '{vacancy.Title}' is not available for editing.");
+            }
```
chrisjensenuk: Wondering if we should move this logic into the client otherwise we'll have to do this do this on every page.  Maybe an optional parameter on GetVacancyForEditAsync or a new function GetDraftVacancyForEditAsync()?

chrisjensenuk: if (vacancy.Status != VacancyStatus.Draft)

sachpatel: There is another way around this if we were to pass a func/action but I don't think that would be favorable.

src/Employer/Employer.Web/Startup.Configure.cs
```diff
@@ -1,21 +1,19 @@
...
 
 namespace Esfa.Recruit.Employer.Web
 {
     public partial class Startup
     {
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
-        public void Configure(IApplicationBuilder app, IHostingEnvironment env,  IOptions<ExternalLinksConfiguration> externalLinks)
+        public void Configure(IApplicationBuilder app, IHostingEnvironment env, IOptions<ExternalLinksConfiguration> externalLinks)
         {
             app.UseStatusCodePagesWithReExecute("/Error/Index/{0}");
```
leewadhams: I think "Index" needs to be removed here else error pages for non 500 errors won't redirect properly to custom error controller. (This issue wasn't introduced in this pull request)

src/Employer/Employer.Web/Orchestrators/SectionsOrchestrator.cs
mgwilliam: fear not... we can move some of this stuff once we have the validation sorted

src/Employer/Employer.Web/Controllers/DashboardController.cs
```diff
@@ -20,10 +20,15 @@ public DashboardController(ILogger<DashboardController> logger, DashboardOrchest
...
-        public async Task<IActionResult> Index()
+        public async Task<IActionResult> Dashboard()
         {
             var employerDetail = (EmployerIdentifier)HttpContext.Items[ContextItemKeys.EmployerIdentifier];
             var vm = await _orchestrator.GetDashboardViewModelAsync(employerDetail);
+            
+            if (TempData.ContainsKey(TempDataKeys.DashboardErrorMessage))
+            {
+                ModelState.AddModelError(string.Empty, TempData[TempDataKeys.DashboardErrorMessage].ToString());
+            }
```
chrisjensenuk: Previously I've done TempData-to-ModleState stuff using an attribute like this. May be worth a butchers. http://piotrgankiewicz.com/2016/04/25/postredirectget-with-new-asp-net-5-mvc-6/

chrisjensenuk: Hmmm. Looks like a mare in Core. Not worth the faff I think

src/Employer/Employer.Web/Startup.Configure.cs
sachpatel: Yep, I spotted this when I was making changes. Will remove.

sachpatel: I have made the amendment. 404 errors (for example) go to the correct action method on the `Error` controller now. The `ConcurrencyException` throw from the orchestrator's for editing a non-draft vacancy goes to the correct action method on the error controller as well.

PR #41, opened 2018-03-08
Some suggested refectorings

leewadhams:
- Jobs to use shared client project.
- Query store to be unaware of query models.
- Jobs project not using QueryViewKeys.cs
- Remove stubs.
- VacancySummaryMapper should be in domain/projections
- Add README for client usage to project.


src/Shared/Recruit.Vacancies.Client/Infrastructure/Client/VacancyClient.cs
```diff
@@ -111,8 +112,17 @@ public VacancyClient(IVacancyRepository repository, IQueryStoreReader reader, IM
         
         public async Task<Dashboard> GetDashboardAsync(string employerAccountId)
         {
-            var key = string.Format(QueryViewKeys.DashboardViewPrefix, employerAccountId);
-            return await _reader.GetDashboardAsync(key);
+            return await _reader.GetDashboardAsync(employerAccountId);
+        }
+
+        public async Task UpdateApprenitceshipProgrammesAsync(IEnumerable<ApprenticeshipProgramme> programmes)
```
sachpatel: Apprenticeships spelling in method name

src/Shared/Recruit.Vacancies.Client/Infrastructure/QueryStore/IQueryStore.cs
```diff
@@ -0,0 +1,12 @@
+using System.Threading.Tasks;
+using Esfa.Recruit.Vacancies.Client.Domain.Projections;
+
+namespace Esfa.Recruit.Vacancies.Client.Infrastructure.QueryStore
+{
+    public interface IQueryStore
+    {
+        Task<T> GetAsync<T>(string key) where T : class, IQueryProjection;
+
+        Task UpdsertAsync<T>(T item) where T : class, IQueryProjection;
```
sachpatel: UpsertAsync spelling error

src/Shared/Recruit.Vacancies.Client/README.md
```diff
@@ -0,0 +1,55 @@
...
+In appSettings.json:
+```json
+{
+    "ConnectionStrings": {
+        "MongoDb": "<replace with mongo connection string>"
+    }
+}
+```
+
+# 3. Instantiate Client
```
mgwilliam: should be h2 ;)

mgwilliam:
namespaces need fixing in the web project
(@leewadhams this is yet another resharper opportunity)

mgwilliam:
look good... just couple of minors:

1. IQueryProjection could be replaced with a base class as there will be several properties that we'll need on all query models (eg. dates and times of creation, model type)

2. ServiceCollectionExtensions namespace is incorrect

3. in future sprints, it's likely we'll want to use to vacancy client from the application classes... so, IVacancyClient.cs should be in the domain not infrastructure

PR #43, opened 2018-03-12
Feature/ER-269 Guid route constraint

sachpatel:
Strengthened route declarations to only capture on actual guids for vacancy id and for guids that are not empty or all 0's.

Put a `TrimModelBinder`


src/Employer/Employer.Web/Configuration/ValidGuidRouteConstraint.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using System;
+
+namespace Esfa.Recruit.Employer.Web.Configuration
+{
+    public class ValidGuidRouteConstraint : IRouteConstraint
```
leewadhams: Can we not use the built in constraint? GuidRouteConstraint

I see that the custom one doesn't accept a Guid.Empty value. Is this important to filter out?

sachpatel: Already making use of the regular in built guid constraint. That one we need to avoid falling into the action on incorrect url specifications. The custom one was just because of the case that a empty or 0 valued guid is acceptable by the regular guid constraint but in our case it would result in a 404, putting the constraint in place allows us to 404 quicker than having to forward the request to the database and resulting in a 404. To be honest, I did this just for learning more about asp.net core and for being defensive. Happy to remove 👍 .

sachpatel:
Replaced by #44

PR #44, opened 2018-03-13
Guid route constraints

sachpatel:


src/Employer/Employer.Web/Controllers/Part1/TitleController.cs
```diff
@@ -24,16 +24,16 @@ public IActionResult Title()
...
         {
             var vm = await _orchestrator.GetTitleViewModelAsync(vacancyId);
             return View(vm);
         }
 
         [HttpPost("create-vacancy", Name = RouteNames.CreateVacancy_Post)]
-        [HttpPost("vacancies/{vacancyId}/title", Name = RouteNames.Title_Post)]
-        public async Task<IActionResult> Title(TitleEditModel m)
+        [HttpPost("vacancies/{vacancyId:guid}/title", Name = RouteNames.Title_Post)]
+        public async Task<IActionResult> Title(TitleViewModel m)
```
chrisjensenuk: This needs to be TitleEditModel

src/Employer/Employer.Web/Controllers/ApplicationProcessController.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Esfa.Recruit.Employer.Web.Controllers
 {
-    [Route("accounts/{employerAccountId}/vacancies/{vacancyId}")]
+    [Route("accounts/{employerAccountId:minlength(6)}/vacancies/{vacancyId:guid}")]
```
chrisjensenuk: This string is repeated a lot. Can/should we put it in a const?

PR #46, opened 2018-03-14
ER-283 training

chrisjensenuk:
Continues from PR #45 which is yet to be reviewed and merged into Master.

- Includes changes outlined in PR #45
- Added Training screen
- Extended `VacancyClient` to return Programmes
- Added `TypeOfDate` attribute to validate ViewModel dates.


mgwilliam:
@chrisjensenuk are we mandating 4-digit year values when the user enters dates? 
it's easier if we do... we had numerous probs on commitments by allowing both 2 and 4 digit years

chrisjensenuk:
I've not been strict on the year format.  I'll do this in the next PR

PR #49, opened 2018-03-16
ER-271 Preview

chrisjensenuk:
- Added Preview section
- Moved Vacancy Programme properties into their own Domain entity
- Added `[Display]` attributes to Enums for human readable text
- When we have 1 associated Organisation then we are not displaying the question
- Added configuration `GoogleMapsPrivateKey`
- Displays the map with a pin.  If Lat & Long are available then map uses these. Otherwise it will display a pin based on the postcode

### DIFF
--- a/src/Employer/Employer.Web/Extensions/DecimalExtensions.cs
+++ b/src/Employer/Employer.Web/Extensions/DecimalExtensions.cs
@@ -1,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Esfa.Recruit.Employer.Web.Extensions
+{
+    public static class DecimalExtensions
+    {
+        public static string AsMoney(this decimal dec)
+        {
+            return $"{dec:N2}".Replace(".00", "");
+        }
+    }
+}
--- a/src/Employer/Employer.Web/Orchestrators/Part1/PreviewOrchestrator.cs
+++ b/src/Employer/Employer.Web/Orchestrators/Part1/PreviewOrchestrator.cs
@@ -1,0 +1,45 @@
+using System;
+using System.Threading.Tasks;
+using Esfa.Recruit.Employer.Web.Extensions;
+using Esfa.Recruit.Employer.Web.ViewModels.Part1.Preview;
+using Esfa.Recruit.Vacancies.Client.Domain;
+using Esfa.Recruit.Vacancies.Client.Domain.Exceptions;
+using Esfa.Recruit.Vacancies.Client.Infrastructure.Client;
+using MongoDB.Bson.Serialization.Serializers;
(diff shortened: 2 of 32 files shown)


src/Employer/Employer.Web/Orchestrators/Part1/PreviewOrchestrator.cs
```diff
@@ -0,0 +1,50 @@
...
+            var vm = new PreviewViewModel
+            {
+                OrganisationName = vacancy.OrganisationName,
+                NumberOfPositions = vacancy.NumberOfPositions?.ToString(),
+                ShortDescription = vacancy.ShortDescription,
+                ClosingDate = vacancy.ClosingDate?.ToString(dateFormat),
+                StartDate = vacancy.StartDate?.ToString(dateFormat),
+                LevelName = vacancy.Programme?.LevelName,
+                Wage = vacancy.Wage?.ToText(),
+                Latitude = "52.4001929857", //temporary stub!!
```
sachpatel: as mentioned, old mother England suggestion!

src/Shared/Recruit.Vacancies.Client/Domain/Enums/ProgrammeLevel.cs
```diff
@@ -0,0 +1,32 @@
...
+        FoundationDegree = 5,
+
+        [Display(Name = "Degree")]
+        Degree = 6,
+
+        [Display(Name = "Masters")]
+        Masters = 7,
+
+        [Display(Name = "Traineeship")]
+        Traineeship = 8
```
sachpatel: Is this one we need to be concerned about? It is not listed on the following page https://www.gov.uk/apprenticeships-guide

src/Employer/Employer.Web/Extensions/DecimalExtensions.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Esfa.Recruit.Employer.Web.Extensions
+{
+    public static class DecimalExtensions
+    {
+        public static string AsMoney(this decimal dec)
```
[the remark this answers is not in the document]
chrisjensenuk: I don't want the £ sign so I'll keep it as N2 for now.

src/Employer/Employer.Web/Orchestrators/Part1/PreviewOrchestrator.cs
chrisjensenuk: Nice idea. I'll look into it

src/Shared/Recruit.Vacancies.Client/Domain/Enums/ProgrammeLevel.cs
chrisjensenuk: Good catch. A cut and paste issue

PR #50, opened 2018-03-16
ER-270 - Resuming a vacancy from the dashboard

chrisjensenuk:
- Nearly all of the functionality is already in place for this ticket
- Explicitly ordering by CreatedDate

### DIFF
--- a/src/Shared/Recruit.Vacancies.Client/Application/Handlers/UpdateDashboardCommandHandler.cs
+++ b/src/Shared/Recruit.Vacancies.Client/Application/Handlers/UpdateDashboardCommandHandler.cs
@@ -24,10 +24,7 @@ namespace Esfa.Recruit.Vacancies.Client.Application.Handlers
         {
             var vacancySummaries = await _repository.GetVacanciesByEmployerAccountAsync<VacancySummary>(message.EmployerAccountId);
             
-            var employerVacancies = await _repository.GetVacanciesByEmployerAccountAsync(message.EmployerAccountId)
-                .OrderBy(v => v.CreatedDate);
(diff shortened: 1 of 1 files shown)


src/Shared/Recruit.Vacancies.Client/Application/Handlers/UpdateDashboardCommandHandler.cs
```diff
@@ -23,7 +23,8 @@ public UpdateDashboardCommandHandler(IQueryStoreWriter writer, IVacancyRepositor
         public async Task Handle(UpdateDashboardCommand message, CancellationToken cancellationToken)
         {
             var employerVacancies = await _repository.GetVacanciesByEmployerAccountAsync(message.EmployerAccountId);
-            var vacancySummaries = employerVacancies.Select(VacancySummaryMapper.MapFromVacancy);
+            var vacancySummaries = employerVacancies.Select(VacancySummaryMapper.MapFromVacancy)
+                .OrderBy(v => v.CreatedDate);
```
sachpatel: Do we have to do this for storage ie. does database need to store the summaries in the specified order? Currently this is being done in the orchestrator/mapping part and it may be more useful there for any other ordering or reverse ordering operations maybe applied from the dashboard view. If this is seen as more important then maybe we remove it from the mapper.

chrisjensenuk: You are right. We are explicitly ordering when we are creating the view model. I just thought we may as well have a 'default' sort order in the storage.

PR #53, opened 2018-03-20
Part1 form restrictions

sachpatel:


mgwilliam:
@sachpatel is the min/max attributes on input fields valid if using novalidate on the form?

src/Employer/Employer.Web/ViewModels/Part1/Training/TrainingViewModel.cs
```diff
@@ -1,12 +1,11 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Linq;
-using System.Threading.Tasks;
 
 namespace Esfa.Recruit.Employer.Web.ViewModels.Part1.Training
 {
     public class TrainingViewModel : TrainingEditModel
     {
+        public int MinYear => DateTime.UtcNow.Year;
```
chrisjensenuk: I'd ignored doing this type of validation in the ViewModel as I think this is a business rule error not a data validation one. But in this case its helpful as we can limit the HTML control. 👍

src/Employer/Employer.Web/Views/Part1/Wage/Wage.cshtml
```diff
@@ -12,13 +12,13 @@
             </h2>
         </div>
         
-        <form asp-route="RouteNames.Wages_Get">
+        <form asp-route="@RouteNames.Wage_Post" novalidate>
```
chrisjensenuk: Good spot!

src/Employer/Employer.Web/Views/Part1/Wage/Wage.cshtml
```diff
@@ -36,7 +36,7 @@
             <div class="form-group">
                 <label asp-for="WeeklyHours" class="form-label-bold">How many hours will the apprentice work each week, including training?</label>
                 <span asp-validation-for="WeeklyHours" class="error-message"></span>
-                <input asp-for="WeeklyHours" class="form-control form-control-1-8" type="number" step="0.01">
+                <input asp-for="WeeklyHours" class="form-control form-control-1-8" type="number" step=".25" min="16" max="48">
```
chrisjensenuk: Is the UI getting too clever here? Should the decision about the hours be done as business rule?  If we can put these limitations in the HTML then maybe we should update the supporting caption to provide the allowable range.

src/Employer/Employer.Web/ViewModels/Part1/Training/TrainingViewModel.cs
sachpatel: This is more of a hint helper. I am just limiting some of the html input fields in a light way. A user can still put 2015 for example. The helper just makes a difference if they are using the up/down arrows.

src/Employer/Employer.Web/Views/Part1/Wage/Wage.cshtml
sachpatel: Again, this is just a light way of restricting values if the user opts to use the up/down arrows. In terms of the 16 and 48, yes you are right, they need to be reflective of the business rule we will validate on server side and maybe we are more informative in the label/caption beforehand.

mgwilliam: 👍

sachpatel:
@mgwilliam 
`is the min/max attributes on input fields valid if using novalidate on the form?` - Yes, they just restrict the lower and upper bounds if using the up/down keys. You can put a higher number if typing in manually.
`also, if the agent falls back to standard input boxes do we have corresponding min/max validation?` - We should do but haven't as of yet. Will look into putting this in the ViewModel validation.

PR #55, opened 2018-03-21
Update for Storage account connections string used by EditVacancyInfo job and queue event

sachpatel:
I don't think you can specify creation of a storage account queue in ARM templates yet. 😫  @EwanNoble would this be okay.

The consuming code creates a queue if it does not already exist so we shouldn't encounter any issues.


src/Employer/Esfa.Recruit.Azure.Resources/azuredeploy.json
```diff
@@ -204,7 +209,8 @@
             },
             "file": {
               "enabled": true
-            }
+            },
+            ""
```
EwanNoble: Can you remove this please 👍

EwanNoble: Can you remove this please 👍

sachpatel: 👍  done

PR #59, opened 2018-03-23
ER-290 Skills & Personal Qualities

chrisjensenuk:
- Skills saved as a list of strings against the vacancy
- The list of default skills come from the Web project's configuration
- There is importance attached to the different columns hence configuration is per column
- Added progressive javascript when adding and removing custom skills. I've just used jQuery to do this for the time being.
- Added a comment to the SkillsOrchestrator detailing a minor complication with the validation which we'll need to handle


src/Employer/Employer.Web/ViewModels/Part2/Skills/SkillsViewModel.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Esfa.Recruit.Employer.Web.ViewModels.Part2.Skills
+{
+    public class SkillsViewModel : SkillsEditModel
+    {
+        public List<SkillViewModel> Column1Checkboxes { get; set; }
+        public List<SkillViewModel> Column2Checkboxes { get; set; }
+        public List<string> CustomSkills { get; set; }
+    }
+
+    public class SkillViewModel
```
sachpatel: Suggestion: Maybe call this `Skill` and move to the Web.Models folder. That way the Skills service doesn't reference the class with ViewModel in the name.

chrisjensenuk:
@leewadhams Initially I didn't persist the checkboxes when adding a custom skill and I found the UX a bit frustrating.  I'd land on the page, click some boxes, add a custom skill and the page would post back and lose the boxes I just selected.  Its easy to change this behaviour so maybe we see how it tests?

src/Employer/Employer.Web/ViewModels/Part2/Skills/SkillsViewModel.cs
chrisjensenuk: I did think about making this service less view centric but as it is UI aware with regards to column1 and column2 I didn't mind this class being bound to the ViewModel.

leewadhams:
@chrisjensenuk Do we need to persist at all when adding a new skill? Just seems a little inconsistent as to what is saved when for non-js version.

1. Check boxes, Add custom skill, click cancel -> Checkboxes and custom skill saved
2. Add custom skill, check boxes, click cancel -> Only custom skill saved.

I wouldn't expect anything to have been saved if I click cancel.

chrisjensenuk:
@leewadhams I see what you mean. I'll have a look 👍

PR #61, opened 2018-03-27
ER-274 Part 2 long desc questions

sachpatel:


src/Employer/Employer.Web/Orchestrators/PreviewOrchestrator.cs
```diff
@@ -32,33 +33,38 @@ public PreviewOrchestrator(IVacancyClient client, IGeocodeImageService mapServic
...
                 CanSubmit = vacancy.CanSubmit,
-                ContactName = string.Empty,
-                ContactEmail = string.Empty,
-                ContactTelephone = string.Empty,
-                ClosingDate = vacancy.ClosingDate.Value,
-                Description = string.Empty,
+                ContactName = vacancy.EmployerContactName,
+                ContactEmail = vacancy.EmployerContactEmail,
+                ContactTelephone = vacancy.EmployerContactPhone,
+                ClosingDate = vacancy.ClosingDate.Value.ToString("dd MMM yyyy"),
```
chrisjensenuk: I don't think we should assume ClosingDate and PossibleStartDate will not be null. I think this page should render even if elements are missing

PR #65, opened 2018-04-03
Validation summary

sachpatel:
Have got the Validation summary working nicely. Have used the approach taken on the MA site. Custom TagHelper is not ideal for generating the type of html we would want to generate so was not used.

Fields with errors have the error message in the summary box have a link to the relevant input field on the form for the user to correct. This is because the `asp-validation-for` tag helper does not generate an html element with the field name in the `id`.

## Remaining Issues
- There is an issue with ClosingDate on the Part 1 Training page, ClosingDate is a calculated field but does not correspond to **one** field on the form.
- The order of the error messages in the validation summary does not correspond to the order of the input items displayed on the form.


sachpatel:
superceded by ER-309 JIRA ticket related PR/branch

PR #68, opened 2018-04-04
ER-335 Refactor Triggering of Dashboard update

leewadhams:
- Removed 'flag' from client methods as to when dashboard should be updated.
- New domain events created for Vacancy.
- New handler that listens for vacancy domain events and updates Dashboard when event is triggered.


src/Shared/Recruit.Vacancies.Client/Application/EventHandlers/UpdateDashboardOnVacancyChange.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        public async Task Handle(VacancyCreatedEvent notification, CancellationToken cancellationToken)
+        {
+            _logger.LogInformation("Handling VacancyCreatedEvent for accountId: {employerAccountId}", notification?.EmployerAccountId);
+            await ReBuildDashboard(notification.EmployerAccountId);
+        }
+
+        public async Task Handle(VacancyUpdatedEvent notification, CancellationToken cancellationToken)
+        {
+            _logger.LogInformation("Handling VacancyUpdatedEvent for accountId: {employerAccountId}", notification?.EmployerAccountId);
```
sachpatel: Can we use `notification.GetType().Name`? also are you missing the $ to do the string interpolation for `employerAccountId`

leewadhams: It's not using string interpolation. The logger uses the string format and the place holder names as the property names. You'll see in them as properties on the log entry in kibana.

sachpatel: 👍

leewadhams: Could do this: 

```$"Handling {notification.GetType().Name} for accountId: {{employerAccountId}}"```

What do you think?

PR #79, opened 2018-04-11
ER-290 ER-291 Validating Qualifications/Skills on Addition/Removal

chrisjensenuk:
History
=====
The bulk of these tickets were done before validations were finalised.  The thinking at the time was validations will be done as part of the Save operation.  The skills/qualifications functionality wanted to build an in-memory list of skills/qualifications before clicking 'Save and continue' and saving them to the DB. This meant we couldn't validate the entries as they were added. We could only do a validation of all skills/qualifications as the vacancy was Saved.

I also wanted to code the page in such a way so we could add progressive javascript enhancement without needing to rewrite the back-end so I added some JS code to demonstrate how this could be done. My intention was to demo this to the dev team for feedback but I never got around to it.  This JS code didn't do any validation either as we are relying on purely server side validations for the time being.

Fast-forward to today. Validation and submitting of a vacancy is now done in two separate operations which means I can now validate on adding/removing of a skill/qualification hence this PR.  I've also removed the JS code.


src/Employer/Employer.Web/Controllers/Part2/SkillsController.cs
```diff
@@ -33,13 +33,6 @@ public SkillsController(SkillsOrchestrator orchestrator)
         [HttpPost("skills", Name = RouteNames.Skills_Post)]
```
leewadhams: Unused using - using System.Collections.Generic;

src/Employer/Employer.Web/Orchestrators/Part2/QualificationsOrchestrator.cs
```diff
@@ -68,27 +68,28 @@ public QualificationsOrchestrator(IVacancyClient client, IOptions<Qualifications
                 throw new ConcurrencyException(string.Format(ErrorMessages.VacancyNotAvailableForEditing, vacancy.Title));
```
leewadhams: Wrong type is being passed in the ILogger<> of the constructor

src/Employer/Employer.Web/Orchestrators/Part2/QualificationsOrchestrator.cs
```diff
@@ -68,27 +68,28 @@ public QualificationsOrchestrator(IVacancyClient client, IOptions<Qualifications
...
-            
+
+            //if we are adding/removing a qualification then just validate and don't persist
+            var validateOnly = m.IsAddingQualification || m.IsRemovingQualification;
+
             return await ValidateAndExecute(vacancy,
                 v => 
                 {
                     var result = _client.Validate(v, ValidationRules);
                     SyncErrorsAndModel(result.Errors, m);
```
leewadhams: Should the m.Qualifications be based on the Vacancy Entity?

```
 //Get the index position for the first invalid qualification
            var qualificationIndex = errors.FirstOrDefault(e => e.PropertyName.StartsWith($"{nameof(m.Qualifications)}["))?.GetIndexPosition();
            if (!qualificationIndex.HasValue)
            {
                return;
            }
```

PR #85, opened 2018-04-16
ER-259 change of state invalid user journey

sachpatel:
On the delete journey, the 'No' path results in a route not found error which will be resolved correctly when #84 is reviewed and merged.


src/Employer/Employer.Web/ViewModels/ErrorMessages.cs
```diff
@@ -3,7 +3,7 @@
     public class ErrorMessages
     {
         public const string VacancyCannotBeViewed = "The vacancy '{0}' can't be viewed.";
-        public const string VacancyNotAvailableForEditing = "The vacancy '{0}' is not available for editing.";
+        public const string VacancyNotAvailableForEditing = "The vacancy '{0}' has been edited by one of your colleagues. Your changes have not been saved.";
```
chrisjensenuk: I think this message makes sense for POSTs but not for GETs.  If this error is triggered during a GET I think the main culprit will be people clicking an old hyperlink/favourite.  Maybe we should keep the "The vacancy '{0}' is not available for editing." message for GETs?

PR #92, opened 2018-04-17
ER-311 Properties to support QA

chrisjensenuk:
- Have shuffled the properties around in the Vacancy entity. Identification and state based properties are now at the top and are `internal set`. The remainder of the fields are alphabetical and `public set`. Helpers are located at the bottom of the class.
- Added MongoDB convention IgnoreIfNullConvention
- We now trap user information when creating, updating, submitting and deleting.

### DIFF
--- a/src/Shared/Recruit.Vacancies.Client/Domain/Entities/SourceType.cs
+++ b/src/Shared/Recruit.Vacancies.Client/Domain/Entities/SourceType.cs
@@ -1,0 +1,9 @@
+﻿namespace Esfa.Recruit.Vacancies.Client.Domain.Entities
+{
+    public enum SourceType
+    {
+        Clone,
+        Extension,
+        New
+    }
+}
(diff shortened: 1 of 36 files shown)


src/Shared/Recruit.Vacancies.Client/Domain/Entities/SourceType.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Esfa.Recruit.Vacancies.Client.Domain.Entities
+{
+    public enum SourceType
+    {
+        Clone,
+        Extension,
+        New
```
sachpatel: maybe `Original` even though ticket says `New`

mgwilliam: @sachpatel the property uses the word "User" because it reflects which user last updated the vacancy (ie. a person)
if there was a system update to the vacancy then this wouldn't be recorded using these field(s)

PR #94, opened 2018-04-19
ER-303 data level authorisation checks & ER-295 Training Provider Navigation Issue Fix

sachpatel:
I have created a class for messages in exceptions called `ExceptionMessages` in the application layer. (This is because that is the deepest layer from which we would raise exceptions, happy to change if necessary).

**Please could somebody dev QA this thoroughly by running it locally and going through all of the pages/actions that interact with a single vacancy. I have tested this multiple times but it would be good for someone to do a fresh round of testing before this ticket goes to QA.**

Here is some guidance:
* Create a new vacancy up to ready for submission ( to the vacancy preview with a filled out vacancy but do not submit)
* Temporarily turn off the `Developer Exception Page` option.
* Visit the different 'Vacancy' pages (to both land on the page and performing a submit action from the page) toggling the value between a correct/unmatching value for the  `employerAccountId` field in the vacancy you are attempting to access/action (using Robo 3T or another Mongo client).
* Submit the vacancy and test landing on the `submitted` page of the process for a vacancy not owned by the employer account you are under.

**Verify landing on the AccessDenied page as well as the logging of the error (visible through the output window or Kibana)**

The following was used for toggling the vacancy data of the vacancy being accessed in Mongo:

```
db.vacancies.updateOne({"title": "test"}, {$set: {"employerAccountId": "{AccountId}"}})

db.vacancies.updateOne({"title": "test"}, {$set: {"employerAccountId": "{AccountIdWithXOnTheEnd}"}}) // incorrect

db.vacancies.updateOne({"title": "test"}, {$set: {"Status": "Referred"}})
db.vacancies.updateOne({"title": "test"}, {$set: {"Status": "Live"}})
db.vacancies.updateOne({"title": "test"}, {$set: {"Status": "Closed"}})
db.vacancies.updateOne({"title": "test"}, {$set: {"Status": "Draft"}})
```


sachpatel:
Hello, I have added a fix for ER-295 which had previously progressed to 'PO Acceptance' but I found an issue with trying to navigate to the confirm training provider page (without the flow through the SelectTrainingProvider page) resulting in a 500 when it should just redirect to the SelectTrainingProvider page. I have added a scenario to the ER-295 ticket `Attempts to bypass SelectTrainingProvider page to land on ConfirmTrainingProvider redirect`

src/Employer/Employer.Web/Orchestrators/Part2/TrainingProviderOrchestrator.cs
```diff
@@ -39,10 +41,20 @@ public TrainingProviderOrchestrator(IEmployerVacancyClient client, ITrainingProv
             return vm;
         }
 
+        public async Task<SelectTrainingProviderViewModel> GetSelectTrainingProviderViewModel(SelectTrainingProviderEditModel m)
+        {
+            var vrm = new VacancyRouteModel { EmployerAccountId = m.EmployerAccountId, VacancyId = m.VacancyId };
```
chrisjensenuk: Don't need this.  Just use SelectTrainingProviderEditModel

src/Employer/Employer.Web/Orchestrators/Part2/VacancyDescriptionOrchestrator.cs
```diff
@@ -39,7 +41,8 @@ public VacancyDescriptionOrchestrator(IEmployerVacancyClient client, ILogger<Vac
 
         public async Task<VacancyDescriptionViewModel> GetVacancyDescriptionViewModelAsync(VacancyDescriptionEditModel m)
         {
-            var vm = await GetVacancyDescriptionViewModelAsync(m.VacancyId);
+            var vrm = new VacancyRouteModel { EmployerAccountId = m.EmployerAccountId, VacancyId = m.VacancyId };
```
chrisjensenuk: ditto

src/Employer/Employer.Web/Controllers/VacancyManageController.cs
```diff
@@ -21,9 +24,12 @@ public VacancyManageController(VacancyManageOrchestrator orchestrator, IEmployer
         }
 
         [HttpGet("", Name = RouteNames.DisplayVacancy_Get)]
-        public async Task<IActionResult> DisplayVacancy([FromRoute]Guid vacancyId)
+        public async Task<IActionResult> DisplayVacancy(VacancyRouteModel vrm)
         {
-            var vacancy = await _client.GetVacancyAsync(vacancyId);
+            var vacancy = await _client.GetVacancyAsync(vrm.VacancyId);
+
+            if (!vacancy.EmployerAccountId.Equals(vrm.EmployerAccountId, StringComparison.OrdinalIgnoreCase))
```
leewadhams: Should this not live in the Orchestrator?

src/Employer/Employer.Web/Orchestrators/Part2/TrainingProviderOrchestrator.cs
sachpatel: This would fall over as it did earlier. It would be calling the method it was already in resulting in the overflow. The signature of the one I mean to call is different and requires a VacancyRouteModel type so is necessary.

src/Employer/Employer.Web/Orchestrators/DeleteVacancyOrchestrator.cs
```diff
@@ -17,16 +19,19 @@ public DeleteVacancyOrchestrator(IEmployerVacancyClient client)
...
         }
 
-        public async Task<DeleteViewModel> GetDeleteViewModelAsync(Guid vacancyId)
+        public async Task<DeleteViewModel> GetDeleteViewModelAsync(VacancyRouteModel vrm)
         {
-            var vacancy = await _client.GetVacancyAsync(vacancyId);
+            var vacancy = await _client.GetVacancyAsync(vrm.VacancyId);
+            
+            if (!vacancy.EmployerAccountId.Equals(vrm.EmployerAccountId, StringComparison.OrdinalIgnoreCase))
+                throw new AuthorisationException(string.Format(ExceptionMessages.VacancyUnauthorisedAccess, vrm.EmployerAccountId, vacancy.EmployerAccountId, vacancy.Title, vacancy.Id));
```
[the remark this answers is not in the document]
sachpatel: I would prefer not to put this in an extension method. I don't like the idea of augmenting the Vacancy entity with this behavior or being concerned with throwing an exception. Leaving it where it is it keeps it open for the orchestrator to apply the check rather than just call it. Also it would be odd when using the Vacancy entity in other consuming apps where this function would not be there if it is an extension in Employer.Web.

src/Employer/Employer.Web/Controllers/ErrorController.cs
```diff
@@ -24,29 +24,52 @@ public ErrorController(ILogger<ErrorController> logger)
...
             if (exceptionFeature != null)
             {
-                var accountId = (string)HttpContext.Items[ContextItemKeys.EmployerIdentifier];
                 string routeWhereExceptionOccurred = exceptionFeature.Path;
                 var exception = exceptionFeature.Error;
 
                 if (exception is InvalidStateException)
                 {
-                    _logger.LogError(exception, $"Exception on path: {routeWhereExceptionOccurred}");                    
+                    _logger.LogError(exception, $"Exception on path: {routeWhereExceptionOccurred}");
```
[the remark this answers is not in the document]
sachpatel: 👍 will look into this.

src/Employer/Employer.Web/Controllers/ErrorController.cs
```diff
@@ -24,29 +24,52 @@ public ErrorController(ILogger<ErrorController> logger)
...
                     TempData.Add(TempDataKeys.DashboardErrorMessage, exception.Message);
                     return RedirectToRoute(RouteNames.Dashboard_Index_Get, new { EmployerAccountId = accountId });
                 }
+
+                if (exception is AuthorisationException)
+                {
+                    _logger.LogError(exception, exception.Message);
```
leewadhams: Should this be logged as a warning as it's not really an application failure and it's a handled scenario?

src/Employer/Employer.Web/Controllers/ErrorController.cs
leewadhams: Add double braces if you want to include routeWhereExceptionOccured as a property on the log message.

src/Employer/Employer.Web/Orchestrators/Part2/TrainingProviderOrchestrator.cs
chrisjensenuk: I see. Ha! Change the name of one for the functions? Cast it?

src/Employer/Employer.Web/Orchestrators/DeleteVacancyOrchestrator.cs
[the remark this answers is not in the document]
chrisjensenuk: OK. There's a lot of repetition and the different ways of working out authorisation is not sitting well with me.  Seems too easy to forget to apply the check.  Wondering if we could have a wrapper class around VacancyClient which has a function `GetVacancyAsync(VacancyId, EmployerAccountId)`. So you can only get a vacancy if you specify the account id too.

src/Employer/Employer.Web/Orchestrators/Part2/TrainingProviderOrchestrator.cs
sachpatel: 👍 Will cast it

src/Employer/Employer.Web/Controllers/ErrorController.cs
sachpatel: 👍

src/Employer/Employer.Web/Controllers/ErrorController.cs
sachpatel: 👍

src/Employer/Employer.Web/Orchestrators/DeleteVacancyOrchestrator.cs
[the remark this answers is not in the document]
sachpatel: I agree about the repetition completely and did try to reduce it. It is still in more than 2 places so it should be put in a common place. Forgetting to call it is also a concern I have had. However I can't think of a better solution and out of the ones suggested, the one in this PR is what I am content with. I have no issue with changing it to a way that the rest of the team agrees with. I will bring it up on Monday.

src/Employer/Employer.Web/Controllers/VacancyManageController.cs
sachpatel: Hi, the reason why I am using the client in the controller is because I need to inspect the status of the vacancy first before branching out into either calling a method that calls redirection or calling another method on the orchestrator which does inspection of the same vacancy. This was all to avoid calling GetVacancy on the client twice or having a method on the orchestrator which calls the client to return a Vacancy to the controller. I can change this if you prefer?

src/Employer/Employer.Web/Orchestrators/DeleteVacancyOrchestrator.cs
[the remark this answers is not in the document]
leewadhams: Might be a controversial one but could we do something like the following: [https://gist.github.com/leewadhams/6d4b8c2d645b5eef4b68eb32690351c8](https://gist.github.com/leewadhams/6d4b8c2d645b5eef4b68eb32690351c8)

src/Employer/Employer.Web/Controllers/VacancyManageController.cs
sachpatel: Have refactored this so the orchestrator is doing most of the work and calling the EmployerClient instead of the controller

src/Employer/Employer.Web/Orchestrators/DeleteVacancyOrchestrator.cs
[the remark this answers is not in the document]
sachpatel: Thanks Lee, I have decided to just plonk the call in a regular static method as then it is in one place and we can make a call later on.

PR #97, opened 2018-04-20
ER-302 QA Website Staff Idams POC

leewadhams:
- Added new skeleton website under QA folder.
- Configured to use WS Federated authentication.
- There's a new user secrets file. Details on Confluence.
- DevOps can use this to start updating the release pipeline for QA Website


leewadhams:
Yep will remove when starting the 'real' screens.

PR #98, opened 2018-04-24
ER-383 Live Vacancy Projection

chrisjensenuk:
- Nothing actually triggers this yet.  To test I temporarily add `await ApproveVacancyAsync(id, user);` to the end of VacancyClient.SubmitVacancyAsync() function.  This will then populate the QueryStore when a vacancy is submitted via the website.
- I've extended the QueryStore to handle deletions - although this ticket doesn't need to do deletions.
- Faking the lat and long for the moment. We'll need to remove this when we add lat/long functionality
- I've made a small tweak to the Submitted view on the website so it looks OK if the VacancyReference is not populated yet.


leewadhams:
Should the Approve event be published to the storage queue and picked up by the job instead of being inline?

src/Shared/Recruit.Vacancies.Client/Infrastructure/QueryStore/Projections/LiveVacancy/WageType.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Esfa.Recruit.Vacancies.Client.Infrastructure.QueryStore.Projections.LiveVacancy
+{
+    public enum WageType
```
leewadhams: Probably one to debate between us but should we just use the Domain enums?

src/Employer/Employer.Web/ViewModels/Submitted/IndexViewModel.cs
```diff
@@ -1,8 +1,12 @@
-﻿namespace Esfa.Recruit.Employer.Web.ViewModels.Submitted
+﻿using Microsoft.EntityFrameworkCore.Query.Expressions;
```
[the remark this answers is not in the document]
chrisjensenuk: Ha! 👍

src/Shared/Recruit.Vacancies.Client/Infrastructure/QueryStore/Projections/LiveVacancy/WageType.cs
chrisjensenuk: Yeah, Sachin raised this yesterday too.  I've spoken to Mark this morning. As the projection is a read-only view of the vacancy then these values are not going to change so having an enum of possible options doesn't really make sense in this context.  So I'm going to just turn the enums into strings instead for the projection.

src/Shared/Recruit.Vacancies.Client/Infrastructure/QueryStore/Projections/LiveVacancy/LiveVacancy.cs
```diff
@@ -0,0 +1,36 @@
...
+        public string EmployerDescription { get; set; }
+        public Address EmployerLocation { get; set; }
+        public string EmployerName { get; set; }
+        public string EmployerWebsiteUrl { get; set; }
+        public int NumberOfPositions { get; set; }
+        public string OutcomeDescription { get; set; }
+        public string ProgrammeId { get; set; }
+        public IEnumerable<Qualification> Qualifications { get; set; }
+        public string ShortDescription { get; set; }
+        public List<string> Skills { get; set; }
```
sachpatel: Could this be IEnumerable or Qualifications become List<Qualification>

src/Shared/Recruit.Vacancies.Client/Infrastructure/QueryStore/Projections/LiveVacancy/TrainingProvider.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Esfa.Recruit.Vacancies.Client.Infrastructure.QueryStore.Projections.LiveVacancy
+{
+    public class TrainingProvider
+    {
+        public long Ukprn { get; set; }
+        public string Name { get; set; }
+        public Address Address { get; set; }
```
sachpatel: The one you are mapping to doesn't map Address so you could probably leave out the property on this class.

src/Shared/Recruit.Vacancies.Client/Infrastructure/QueryStore/Projections/LiveVacancy/LiveVacancy.cs
chrisjensenuk: 👍 I'll change to IEnumerable

PR #101, opened 2018-04-26
VacancyClient refactor

chrisjensenuk:
When it comes to functions that send a command the VacancyClient should only be a skinny wrapper around the command creation.  There was extra logic in VacancyClient that should really be in the handlers.  So I've moved this logic into the relevant handler.

I've also removed some redundant code that will never be executed due to the InvalidStateException checks that we've introduced.

### DIFF
--- a/src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/SubmitVacancyCommandHandler.cs
+++ b/src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/SubmitVacancyCommandHandler.cs
@@ -5,6 +5,8 @@ using Esfa.Recruit.Vacancies.Client.Domain.Repositories;
 using MediatR;
 using System.Threading;
 using System.Threading.Tasks;
+using Esfa.Recruit.Vacancies.Client.Domain.Entities;
+using Esfa.Recruit.Vacancies.Client.Domain.Services;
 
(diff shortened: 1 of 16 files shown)


src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/SubmitVacancyCommandHandler.cs
```diff
@@ -5,29 +5,48 @@
...
         {
             _repository = repository;
             _messaging = messaging;
+            _timeProvider = timeProvider;
         }
 
         public async Task Handle(SubmitVacancyCommand message, CancellationToken cancellationToken)
         {
-            await _repository.UpdateAsync(message.Vacancy);
+            var vacancy = await _repository.GetVacancyAsync(message.VacancyId);
```
leewadhams: I guess the downside of this change is that we have to Get the vacancy more than once in the request.

chrisjensenuk: Yep. I spoke to Mark about this. Its a necessary performance hit as we can't be sure when the command is going to be executed (in the future this may be done out of process) so the potential for overwrites increases if we keep a handle to the full vacancy.  This way the editing time is shorter so the risk of overwrites diminishes.

PR #110, opened 2018-05-09
Fe changes

sdholder:
Created new step by step part 1 pages, made Preview the same as FAA and added error page styling


src/Employer/Employer.Web/Views/Shared/_Part1Header.cshtml
```diff
@@ -56,7 +49,7 @@
         </div>
     </div>
 </div>
-<hr>
+<br><br>
```
chrisjensenuk: Should we get .example-advert to add this margin?

sdholder: Problem is, when the example advert is hidden, the margin will no longer be there and we need it when the example div is hidden as well.  We could create an empty element with css margin-bottom after the sample, but if we are going to do this might as well use a <br>.

PR #111, opened 2018-05-10
ER-350 Log reviewer

leewadhams:
Update to ER-350 to log the user id of the staff idams user when the review is started.

### DIFF
--- a/src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/StartVacancyReviewCommandHandler.cs
+++ b/src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/StartVacancyReviewCommandHandler.cs
@@ -3,27 +3,35 @@ using Esfa.Recruit.Vacancies.Client.Domain.Repositories;
 using MediatR;
 using System.Threading;
 using System.Threading.Tasks;
-using System;
 using Esfa.Recruit.Vacancies.Client.Domain.Entities;
-using Esfa.Recruit.Vacancies.Client.Domain.Messaging;
-using Esfa.Recruit.Vacancies.Client.Domain.Events;
+using Esfa.Recruit.Vacancies.Client.Domain.Services;
+using Esfa.Recruit.Vacancies.Client.Domain.Exceptions;
 
 namespace Esfa.Recruit.Vacancies.Client.Application.CommandHandlers
 {
     public class StartVacancyReviewCommandHandler: IRequestHandler<StartVacancyReviewCommand>
     {
         private readonly IVacancyReviewRepository _vacancyReviewRepository;
+        private readonly ITimeProvider _time;
 
-        public StartVacancyReviewCommandHandler(IVacancyReviewRepository vacancyReviewRepository)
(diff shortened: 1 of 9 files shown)


src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/StartVacancyReviewCommandHandler.cs
```diff
@@ -0,0 +1,31 @@
...
+        private readonly IVacancyReviewRepository _vacancyReviewRepository;
+
+        public StartVacancyReviewCommandHandler(IVacancyReviewRepository vacancyReviewRepository)
+        {
+            _vacancyReviewRepository = vacancyReviewRepository;
+        }
+
+        public async Task Handle(StartVacancyReviewCommand message, CancellationToken cancellationToken)
+        {
+            var review = await _vacancyReviewRepository.GetAsync(message.ReviewId);
```
chrisjensenuk: Should we check the status of the review just in case?

src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/StartVacancyReviewCommandHandler.cs
```diff
@@ -0,0 +1,31 @@
...
+            review.Status = ReviewStatus.UnderReview;
+            review.ReviewedByUserId = message.UserId;
```
chrisjensenuk: Should we also trap the datetime too?

src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/StartVacancyReviewCommandHandler.cs
leewadhams: Yep good point.

src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/StartVacancyReviewCommandHandler.cs
leewadhams: And another good point 👍

src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/StartVacancyReviewCommandHandler.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            var review = await _vacancyReviewRepository.GetAsync(message.ReviewId);
+
+            if (review.Status != ReviewStatus.PendingReview)
+            {
+                throw new InvalidStateException($"Review is not in correct state to Start the review. Current State: {review.Status}");
+            }
+
+            review.Status = ReviewStatus.UnderReview;
+            review.ReviewedByUserId = message.UserId;
```
sachpatel: Would we not want to store the email address/name of the user doing the review similar to how we are for the CreatedBy, LastUpdatedBy and SubmittedBy on employer web maybe.

PR #113, opened 2018-05-10
ER-339 - Updating recruitment footer links

sachpatel:


src/Employer/Employer.sln
```diff
@@ -19,7 +19,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Recruit.Vacancies.Client",
 EndProject
```
chrisjensenuk: Do you mean to change this file?

PR #118, opened 2018-05-15
ER-440 Add shard keys for collections

leewadhams:
- queryView & vacancyReviews effected.
- Will need to delete the collections from the environments before deployment.


leewadhams:
We will do but there may be more be more effort in doing that one as we may need to create a specific shard key field. There's a ticket in the backlog to address it after private beta release as the impact will still be relatively low with the volume of vacancies in private beta.

PR #122, opened 2018-05-16
Update status shown to the Employer

leewadhams:
Employer shown different status to that of the underlying vacancy


leewadhams:
👍 Updated

PR #131, opened 2018-05-18
ER-404 and ER-418 error page and feedback links updates

sachpatel:
Trying to keep logic out of views where possible. Haven't used/created another layout page as there is no real need for it (yet).


sachpatel:
Could have actually put the piece of the feedback banner selection and logic in a ViewComponent which has access to ViewContext, ViewData but decided against that as I think this is a temporary feature.

PR #133, opened 2018-05-22
Docker compose change and env config

sachpatel:
Hi, I don't think the changes in this pull request would adversely affect anyone else running the solutions and using Docker for local dependencies. For me it will make a difference as it will allow me to use the local Redis container dependencies for logging without having to change appSettings.json constantly and having to exclude committing it.

Also the `docker-compose.yaml` change makes it easier for me to run an elastic v5 + container without configuring the virtual page memory on the virtual Docker host frequently. Elastic search requires 4GB.

### DIFF
--- a/src/Employer/Employer.Web/Program.cs
+++ b/src/Employer/Employer.Web/Program.cs
@@ -30,14 +30,6 @@ namespace Esfa.Recruit.Employer.Web
                 .UseKestrel(c => c.AddServerHeader = false)
                 .UseStartup<Startup>()
                 .UseUrls("http://localhost:5020")
-                .ConfigureAppConfiguration((hostingContext, config) =>
-                    {
-                        var hostingEnvironment = hostingContext.HostingEnvironment;
-                        if (hostingEnvironment.IsDevelopment())
-                        {
-                            config.AddUserSecrets<Startup>();
-                        }
-                    })
                 .UseNLog()
                 .Build();
     }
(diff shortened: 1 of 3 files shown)


src/Employer/Employer.Web/Program.cs
```diff
@@ -36,6 +36,7 @@ public static void Main(string[] args)
                         if (hostingEnvironment.IsDevelopment())
                         {
                             config.AddUserSecrets<Startup>();
+                            config.AddEnvironmentVariables();
```
leewadhams: I think this is redundant (I also think AddUserSecrets is redundant also) as it sounds like it's setup by the fact we use .CreateDefaultBuilder()

src/QA/QA.Web/Program.cs
```diff
@@ -76,6 +76,7 @@ private static X509Certificate2 BuildCertificate()
                         if (hostingEnvironment.IsDevelopment())
                         {
                             config.AddUserSecrets<Startup>();
+                            config.AddEnvironmentVariables();
```
leewadhams: Same as above.

src/Employer/Employer.Web/Program.cs
leewadhams: I think this is redundant (I also think AddUserSecrets is redundant also) as it sounds like it's setup by the fact we use .CreateDefaultBuilder()

src/QA/QA.Web/Program.cs
leewadhams: Same as above

PR #135, opened 2018-05-22
ER-406 Logging

leewadhams:
Concentrated adding logging in the following area:
- Jobs
- Error Handlers
- Command Handlers
- Event Handler

There is certainly more we can log and improve what we log. If there's anything you think is essential for private beta let me know.


src/Employer/Employer.Web/Orchestrators/Part1/EmployerOrchestrator.cs
```diff
@@ -105,5 +106,16 @@ private LocationOrganisationViewModel MapLegalEntitiesToOrgs(LegalEntity data)
         {
             return new LocationOrganisationViewModel { Id = data.LegalEntityId.ToString(), Name = data.Name };
         }
+        
+        private IEnumerable<LocationOrganisationViewModel> BuildLegalEntityViewModels(EditVacancyInfo info, string employerAccountId)
+        {
+            if (info == null || !info.LegalEntities.Any())
+            {
+                Logger.LogError("No legal entities found for {employerAccountId}", employerAccountId);
+                return null; // TODO: Can we carry on without a list of legal entities.
```
sachpatel: How about we do something on the start of the session where we check the user is associated to an employer account with at least one legal entity. If there are none then we redirect them to specific page (not 403) instructing them to add legal entities. Not strictly associated with this bit of code but something to prevent the case. Chris has made some changes to do with starting sessions on #134 .

src/Jobs/Recruit.Vacancies.Jobs/EditVacancyInfo/EditVacancyInfoUpdater.cs
```diff
@@ -1,22 +1,28 @@
...
+        public EditVacancyInfoUpdater(IJobsVacancyClient client, ILogger<EditVacancyInfoUpdater> logger)
         {
             _client = client;
+            _logger = logger;
         }
 
         internal async Task UpdateEditVacancyInfo(string employerAccountId)
         {
-            var legalEntities = await _client.GetEmployerLegalEntitiesAsync(employerAccountId);
+            var legalEntities = (await _client.GetEmployerLegalEntitiesAsync(employerAccountId)).ToList();
```
sachpatel: Don't think ToList is needed.

src/Jobs/Recruit.Vacancies.Jobs/VacancyStatus/LiveVacancyStatusInspector.cs
```diff
@@ -22,16 +21,20 @@ public LiveVacancyStatusInspector(ILogger<LiveVacancyStatusInspector> logger, IJ
 
         internal async Task InspectAsync()
         {
-            var vacancies = await _client.GetLiveVacancies();
-
+            var vacancies = (await _client.GetLiveVacancies()).ToList();
+            int numberClosed = 0;
+            
             foreach (var vacancy in vacancies)
```
sachpatel: `foreach (var vacancy in vacancies.Where(v => v.ClosingDate <= _timeProvider.Now))` instead of the if condition being called in the loop maybe.

src/Jobs/Recruit.Vacancies.Jobs/VacancyStatus/VacancyStatusJob.cs
```diff
@@ -4,7 +4,7 @@
 using Microsoft.Azure.WebJobs;
 using Microsoft.Extensions.Logging;
 
-namespace Esfa.Recruit.Vacancies.Jobs.ApprenticeshipProgrammes
```
sachpatel: 🤐

src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/LiveVacancyCommandHandler.cs
```diff
@@ -7,28 +7,38 @@
...
         {
+            _logger = logger;
             _repository = repository;
             _messaging = messaging;
             _timeProvider = timeProvider;
         }
 
         public async Task Handle(LiveVacancyCommand message, CancellationToken cancellationToken)
         {
+            _logger.LogInformation("Setting vacancy Live for vacancy {vacancyReference}.", message.VacancyId);
```
sachpatel: `vacancyId` :( I know what you mean, it is confusing using Id's and referencenumbers in different places. Should we just stick to using VacancyId, because at least that is always there. It mean we will have ti do a lookup when we are investigating/stepping through. I know this is to do with the QA process where we surface the reference number (not the id)

src/Jobs/Recruit.Vacancies.Jobs/VacancyEvents/VacancyEventsJob.cs
```diff
@@ -28,9 +28,9 @@ public VacancyEventsJob(ILogger<VacancyEventsJob> logger, VacancyEventHandler va
                 
                 await UnpackAndExecute(eventItem.EventType, eventItem.Data);
             }
-            catch (Exception ex)
+            catch (JsonException ex)
```
chrisjensenuk: Worth adding a fallback catch Exception too?

src/Jobs/Recruit.Vacancies.Jobs/VacancyStatus/LiveVacancyStatusInspector.cs
chrisjensenuk: I don't mind this foreach. There may be other processing we want to do in future for other live vacancies

src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/ApproveReferredVacancyReviewCommandHandler.cs
```diff
@@ -3,46 +3,57 @@
...
             _vacancyReviewRepository = vacancyReviewRepository;
             _vacancyRepository = vacancyRespository;
             _messaging = messaging;
         }
 
         public async Task Handle(ApproveReferredVacancyReviewCommand message, CancellationToken cancellationToken)
         {
+            _logger.LogInformation("Approving referred review {reviewId} for vacancy {vacancyReference}.", message.ReviewId, message.Vacancy.VacancyReference);
+            
             var vacancy = await _vacancyRepository.GetVacancyAsync(message.Vacancy.Id);
```
chrisjensenuk: I think we should change the ApproveReferredVacancyReviewCommand so it only takes a VacancyId and not a full Vacancy.

src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/ApproveReferredVacancyReviewCommandHandler.cs
```diff
@@ -3,46 +3,57 @@
...
+            if (!review.CanApprove)
+            {
+                _logger.LogWarning($"Unable to approve review {{reviewId}} for vacancy {{vacancyReference}} due to review having a status of {review.Status}.", message.ReviewId, vacancy.VacancyReference);
+                return;
+            }
 
             review.ManualOutcome = ManualQaOutcome.Approved;
             review.Status = ReviewStatus.Closed;
 
+            await _vacancyRepository.UpdateAsync(message.Vacancy);
```
chrisjensenuk: Do we need to do this I don't think it has changed? If we need to update I think we should be passing in vacancy not message.Vacancy

src/Shared/Recruit.Vacancies.Client/Application/EventHandlers/RePublishEventToEventStore.cs
```diff
@@ -4,19 +4,24 @@
...
-                                                   INotificationHandler<VacancySubmittedEvent>,
-                                                   INotificationHandler<VacancyReviewApprovedEvent>,
-                                                   INotificationHandler<VacancyReviewReferredEvent>
+    public class RePublishEventToEventStore : 
+                                            INotificationHandler<VacancyCreatedEvent>,
+                                            INotificationHandler<VacancyDraftUpdatedEvent>,
+                                            INotificationHandler<VacancySubmittedEvent>,
+                                            INotificationHandler<VacancyReviewApprovedEvent>,
+                                            INotificationHandler<VacancyReviewReferredEvent>,
+                                            INotificationHandler<UserSignedInEvent>
```
chrisjensenuk: In ER-337 (I've just merged) I've re-purposed the UserSignedInEvent so it doesn't need to be republished.  I have added a SetupEmployerEvent and a SetupEmployerEventHandler instead.  Could you remove SetupEmployerEventHandler please and add the handling to this class instead?

src/Employer/Employer.Web/Orchestrators/Part1/EmployerOrchestrator.cs
leewadhams: There will always be at least one Organisation when querying the account api. This scenario is more if the querystore hasn't been updated with that information.

src/Jobs/Recruit.Vacancies.Jobs/EditVacancyInfo/EditVacancyInfoUpdater.cs
leewadhams: It is enumerated multiple times isn't it?

src/Jobs/Recruit.Vacancies.Jobs/VacancyStatus/LiveVacancyStatusInspector.cs
leewadhams: 👍

src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/LiveVacancyCommandHandler.cs
leewadhams: 👍

src/Jobs/Recruit.Vacancies.Jobs/VacancyEvents/VacancyEventsJob.cs
leewadhams: I put generic exception handling into the Handler class. It made it easier to log the type of event we were handling. This handler just catches any deserialisation exceptions.

src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/ApproveReferredVacancyReviewCommandHandler.cs
leewadhams: Where would you do the update of the vacancy?

src/Jobs/Recruit.Vacancies.Jobs/VacancyEvents/VacancyEventsJob.cs
chrisjensenuk: 👍

src/Shared/Recruit.Vacancies.Client/Application/EventHandlers/RePublishEventToEventStore.cs
leewadhams: 👍

PR #136, opened 2018-05-23
ER-499 job to rebuild employer dashboard

sachpatel:
This is a very basic job. I have used an inner class for the message at the moment as it is not event driven and just to keep it simple. This is a job that will change soon after private beta.

An example message to be published manually onto the queue can be:
```
{
"EmployerAccountId": "MYJR4X"
}
```

### DIFF
--- a/src/Jobs/Recruit.Vacancies.Jobs/DashboardGenerator/DashboardCreator.cs
+++ b/src/Jobs/Recruit.Vacancies.Jobs/DashboardGenerator/DashboardCreator.cs
@@ -1,0 +1,20 @@
+﻿using Esfa.Recruit.Vacancies.Client.Infrastructure.Services;
+using System.Threading.Tasks;
+
+namespace Esfa.Recruit.Vacancies.Jobs.DashboardGenerator
+{
+    public class DashboardCreator
+    {
+        private readonly IDashboardService _dashboardService;
+
+        public DashboardCreator(IDashboardService dashboardService)
+        {
+            _dashboardService = dashboardService;
+        }
+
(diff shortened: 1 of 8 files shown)


src/Jobs/Recruit.Vacancies.Jobs/DashboardGenerator/DashboardCreator.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Esfa.Recruit.Vacancies.Client.Infrastructure.Services;
+using System.Threading.Tasks;
+
+namespace Esfa.Recruit.Vacancies.Jobs.DashboardGenerator
+{
+    public class DashboardCreator
```
chrisjensenuk: Do we need this class? Should/can we just inject the IDashboardService straight into DashboardGeneratorJob?

src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/DashboardService.cs
```diff
@@ -0,0 +1,44 @@
...
+        private readonly IQueryStoreWriter _queryStoreWriter;
+
+        public DashboardService(IVacancyRepository repository, IQueryStoreWriter queryStoreWriter, ILogger<DashboardService> logger)
+        {
+            _logger = logger;
+            _repository = repository;
+            _queryStoreWriter = queryStoreWriter;
+        }
+
+        public async Task ReBuildDashboard(string employerAccountId)
```
chrisjensenuk: ReBuildDashboardAsync?

PR #138, opened 2018-05-23
ER-504 display 404 for unfound vacancies

sachpatel:


src/QA/QA.Web/Controllers/ErrorController.cs
```diff
@@ -46,6 +47,9 @@ public IActionResult ErrorHandler()
                     case NotFoundException _:
                         _logger.LogError(exception, $"Exception on path: {{routeWhereExceptionOccurred}}", exceptionFeature.Path);
                         return View(ViewNames.ErrorView, GetViewModel(HttpStatusCode.NotFound));
+                    case VacancyNotFoundException _:
+                        _logger.LogError(exception, exception.Message);
```
chrisjensenuk: I don't think this should be logged as an error.

sachpatel: I think it should. Want to be able to see if we get any of those NullReferenceExceptions like on the test environment and don't want it mixed with warnings. Will probably change to Warning level after a week or two of use.

PR #144, opened 2018-05-25
ER-508  - Strengthening the url validation

sachpatel:
:boom: Have put comment in code ☹️ just to break down the rules for the url validation.


mgwilliam:
relative would mean it's relative to FAA

PR #146, opened 2018-05-29
ER-515 renaming cookies and turning off arr affinity cookie

sachpatel:
One thing I noticed after while doing this PR and giving cookies unique names between the two sites is I could run the employer app and the qa app in the same browser without conflicts in cookie resolution.

### DIFF
--- a/src/Employer/Employer.Web/Configuration/CookieNames.cs
+++ b/src/Employer/Employer.Web/Configuration/CookieNames.cs
@@ -1,7 +1,10 @@
-﻿namespace Esfa.Recruit.Employer.Web.Configuration
+namespace Esfa.Recruit.Employer.Web.Configuration
 {
     public static class CookieNames
     {
         public const string ShowExampleVacancy = "showExampleVacancy";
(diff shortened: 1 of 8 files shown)


src/Employer/Employer.Web/Configuration/CookieNames.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace Esfa.Recruit.Employer.Web.Configuration
+{
+    public static class CookieNames
+    {
+        public const string RecruitData = "recruit-data";
+        public const string RecruitTempData = "recruit-temp-data";
+        public const string AntiForgeryCookie = "recruit-x-csrf";
+        public const string SetupEmployer = "setup_employer_{0}";
```
chrisjensenuk: Change the underscores to hyphens if you want. I only chose underscores as that that's what the "seen_cookie_message" is using.

sachpatel: 👍

PR #152, opened 2018-05-30
ER-495 skills from querystore

chrisjensenuk:
I'm getting the skills from Mongo in the constructor of the SkillsOrchestrator.  Do you think this should be cached aside in memory?

Additional refactor - renamed VacancyClient.SetupEmployer to SetupEmployerAsync


src/Employer/Employer.Web/Orchestrators/Part2/SkillsOrchestrator.cs
```diff
@@ -1,35 +1,35 @@
...
+        private readonly CandidateSkills _candidateSkills;
 
-        public SkillsOrchestrator(IEmployerVacancyClient client, IOptions<SkillsConfiguration> skillsConfigOptions, ILogger<SkillsOrchestrator> logger) : base(logger)
+        public SkillsOrchestrator(IEmployerVacancyClient client, ILogger<SkillsOrchestrator> logger) : base(logger)
         {
             _client = client;
-            _skillsConfig = skillsConfigOptions.Value;
-        }
-
+            _candidateSkills = _client.GetCandidateSkillsAsync().Result;
```
sachpatel: I don't think this is a good idea. Can you use a Lazy property that way it only gets populated when it is called the first time.

src/Data/CosmosDb/Console.RecruitSeedDataWriter/Data/CandidateSkills.json
```diff
@@ -1,6 +1,6 @@
 ﻿{
   "_id": "CandidateSkills",
-  "Skills": [
+  "column1Skills": [
```
sachpatel: Could this document not just be as it was.  don't think the database representation should know anything about how it is to be presented in the UI. Maybe you can split the list into two lists in the orchestrator?

src/Employer/Employer.Web/Orchestrators/Part2/SkillsOrchestrator.cs
chrisjensenuk: Its always used at least once per request so doing this in constructor is OK. I did initially look at trying to lazyload this but it added extra complexity without enough gain to my mind.

sachpatel: suggestion:

PR #153, opened 2018-05-30
Refactored utility to not depend on vacancyclient

sachpatel:
This doesn't need to be approved or merged unless people are happy with it.

This was just trying to move the responsibility of the `Utility` class we are using in `Employer.Web` to not have the `VacancyClient` dependency passed in and not having to know about it. Because the `Utility` class has methods we are using for dealing with cross cutting checks at the orchestrator level for Vacancy objects it seems one way to kind of decouple this si to use a function that returns a vacancy instead.


sachpatel:
Will turn the `Utility` class into a non-static buddy class that can be injected and used by the Vacancy orchestrator/controllers making it mock-able and testable without causing the orchestrator and calling places themselves needing more setup that is typically introduced by using static class for side effects.

PR #157, opened 2018-06-06
Hotfix shard issue

leewadhams:
Queries failing against mongo due to sorting across shards.

Moved the sorting into .Net code.

### DIFF
--- a/src/Shared/Recruit.Vacancies.Client/Infrastructure/Repositories/MongoDbVacancyRepository.cs
+++ b/src/Shared/Recruit.Vacancies.Client/Infrastructure/Repositories/MongoDbVacancyRepository.cs
@@ -11,6 +11,7 @@ using System.Linq.Expressions;
 using Esfa.Recruit.Vacancies.Client.Application.Exceptions;
 using Esfa.Recruit.Vacancies.Client.Infrastructure.Exceptions;
 using Microsoft.Extensions.Logging;
+using Polly;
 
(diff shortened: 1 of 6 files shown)


src/Shared/Recruit.Vacancies.Client/Infrastructure/Repositories/MongoDbVacancyRepository.cs
```diff
@@ -80,22 +81,20 @@ public MongoDbVacancyRepository(ILogger<MongoDbVacancyRepository> logger, IOptio
...
-
             var collection = GetCollection<Vacancy>();
-            var result = await RetryPolicy.ExecuteAsync(() => collection.FindAsync(filter, options));
-            return result.SingleOrDefault();
+            var result = await RetryPolicy.ExecuteAsync(context => collection.FindAsync(filter), new Context(nameof(GetSingleVacancyForPostcode)));
+            
+            return result
+                        .ToList()
+                        .OrderByDescending(x => x.VacancyReference)
+                        .SingleOrDefault();
```
chrisjensenuk: This needs to be FirstOrDefault as we now have many vacancies returning. Unless we reintroduce FindOptions Limit = 1.

mgwilliam: good spot!

PR #163, opened 2018-06-08
ER-315 Create online application vacancy

sachpatel:
Have added a feature toggle with the default false called `AllowThroughFaaApplicationMethod` in employer web. Have not used a third party library as for the moment we only need it to control based on a boolean value.

Have also added the `ApplicationMethod` enum which also sets the `ApplicationMethod` in the `LiveVacancy` query view when a vacancy is approved.

There are validation tests around the ApplicationMethod as well as tests around some logic introduced in the `ApplicationProcessOrchestrator`

Have tested the Non-JS and JS version of the application process page with radio buttons for selecting the application method (if the feature switch is toggled to `true`).

Have updated the `VacancyPreview` view and all other views that would display the application method.

In order to backfill `LiveVacancy` and `vacancy` documents already stored in the database the following snippets ca be run and will need to be run on each environment before this code is moved onto them.

```
db.queryViews.updateMany(
{"type":"LiveVacancy", "applicationMethod": {$exists:false}},
{$set:{"applicationMethod": "ThroughExternalApplicationSite"}}, 
false, 
true
)
```

```
db.vacancies.updateMany({"applicationMethod": {$exists:false}}, 
                    {$set:{"applicationMethod": "ThroughExternalApplicationSite"}}, false, true);
```

Will have a word with Ops about incorporating the feature toggle switch into VSTS setting replacement.


src/QA/QA.Web/Views/Review/Review.cshtml
```diff
@@ -267,22 +239,38 @@
...
-
+    <div class="grid-row">
+        <div class="column-full">
+            <h2 class="heading-large">Application process</h2>
+        </div>
+    </div>
+    <div class="grid-row">
+        <div class="column-full">
+            <p asp-condition="@Model.HasSpecifiedThroughFaaApplicationMethod">Candidates will apply through the find an apprenticeship service</p>
+            <div asp-condition="@Model.HasApplicationUrl">
```
chrisjensenuk: Should we add HasSpecifiedThroughExternalApplicationMethod to QA too rather than this assumption?

src/Shared/Recruit.Vacancies.Client/Application/Validation/Fluent/FluentVacancyValidator.cs
```diff
@@ -389,6 +386,42 @@ private void ValidateOutcomeDescription()
...
+                        .WithMessage("You must select an application method")
+                        .WithErrorCode("85")
+                    .RunCondition(VacancyRuleSet.ApplicationMethod)
+                    .WithRuleId(VacancyRuleSet.ApplicationMethod);
+
+            When(x => x.ApplicationMethod == ApplicationMethod.ThroughFindAnApprenticeship, () =>
+            {
+                RuleFor(x => x.ApplicationUrl)
+                    .Empty()
+                        .WithMessage("The application url must be empty when application through external website option is specified")
```
chrisjensenuk: Should this be "...through Find an Apprenticeship is specified" ?

src/Shared/Recruit.Vacancies.Client/Application/Validation/Fluent/FluentVacancyValidator.cs
```diff
@@ -389,6 +386,42 @@ private void ValidateOutcomeDescription()
...
+                        .WithErrorCode("86")
+                    .RunCondition(VacancyRuleSet.ApplicationMethod)
+                    .WithRuleId(VacancyRuleSet.ApplicationMethod);
+
+                RuleFor(x => x.ApplicationInstructions)
+                    .Empty()
+                        .WithMessage("The application process must be empty when application through external website option is specified")
```
chrisjensenuk: ditto

leewadhams:
Is the query to update the QueryViews collection going to work in Cosmos as we aren't specifying the shard key?

PR #165, opened 2018-06-11
Er 277 character count

sdholder:
Added functionality to count characters in various text fields and text-area fields

### DIFF
--- a/src/Employer/Employer.Web/wwwroot/javascripts/application.js
+++ b/src/Employer/Employer.Web/wwwroot/javascripts/application.js
@@ -104,4 +104,38 @@ $(function() {
     sfa.navigation.init();
 });
 
-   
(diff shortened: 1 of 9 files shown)


src/Employer/Employer.Web/wwwroot/javascripts/application.js
```diff
@@ -104,4 +104,38 @@ $(function() {
     sfa.navigation.init();
 });
 
-   
\ No newline at end of file
+/* -----------------------
+Character count behaviour
+-------------------------- */
+characterCount = function(n) {
+  var i = $(n)
```
sachpatel: 👨‍🔬 bit of a science equation here. Could we have slightly more meaningful names to the variables. Or has this been lifted from a  library.

PR #174, opened 2018-06-26
ER-316 successful application

chrisjensenuk:
- Added command `ApplicationReviewSuccessfulCommand`
- Added command handler `ApplicationReviewCommandHandler` to handle application review commands
- Added event `ApplicationReviewSuccessfulEvent`

After setting application review to 'Successful' an `ApplicationReviewSuccessfulEvent` is published.
This event is handled by:
- `UpdateDashboardOnVacancyChange` to rebuild the dashboard query view (there are new application counts against the vacancy that need updating)
- `UpdateVacancyApplicationsOnApplicationReviewChange` - to rebuild the applications against the vacancy
- In a future ticket we'll add an event handler to push a message to FAA to update it's application

### DIFF
--- a/src/Employer/Employer.Web/Extensions/ApplicationReviewExtensions.cs
+++ b/src/Employer/Employer.Web/Extensions/ApplicationReviewExtensions.cs
@@ -37,7 +37,7 @@ namespace Esfa.Recruit.Employer.Web.Extensions
                 Improvements = r.Application.Improvements,
                 Outcome = r.Status,
                 Phone = r.Application.Phone,
-                Qualifcations = r.Application.Qualifications?.Select(q =>
+                Qualifications = r.Application.Qualifications?.Select(q =>
                     new QualificationViewModel
                     {
                         Grade = q.Grade,
--- a/src/Employer/Employer.Web/Extensions/ApplicationReviewExtensions.cs
+++ b/src/Employer/Employer.Web/Extensions/ApplicationReviewExtensions.cs
@@ -35,6 +35,7 @@ namespace Esfa.Recruit.Employer.Web.Extensions
                 EducationToYear = r.Application.EducationToYear,
                 HobbiesAndInterests = r.Application.HobbiesAndInterests,
                 Improvements = r.Application.Improvements,
+                Outcome = r.Status,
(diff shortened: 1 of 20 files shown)


src/Employer/Employer.Web/Views/ApplicationReview/ApplicationReview.cshtml
```diff
@@ -253,21 +225,18 @@
...
-                </div>
 
-            </div>
+        <section>
+            <h2 class="heading-large">Skills the Employer wants</h2>
 
             <p asp-condition="@Model.HasNoSkills">I don't have the skills the employer wants</p>
 
             <p asp-condition="@Model.HasSkills">
                 @Html.Raw(string.Join(", ", Model.Skills.Select(s => string.Format("<span>{0}</span>", Html.Encode(s)))))
```
sachpatel: Do we have to surround each skill in a `span`? 👍 for the encoding, I think we should do this on the preview views as well.

src/Employer/Employer.Web/Extensions/ApplicationReviewExtensions.cs
```diff
@@ -35,6 +35,7 @@ public static ApplicationReviewViewModel ToViewModel(this ApplicationReview r)
                 EducationToYear = r.Application.EducationToYear,
                 HobbiesAndInterests = r.Application.HobbiesAndInterests,
                 Improvements = r.Application.Improvements,
+                Outcome = r.Status,
                 Phone = r.Application.Phone,
                 Qualifcations = r.Application.Qualifications?.Select(q =>
```
sachpatel: Qualifications

src/Employer/Employer.Web/Views/ApplicationReview/ApplicationReview.cshtml
chrisjensenuk: 👍

src/Employer/Employer.Web/Extensions/ApplicationReviewExtensions.cs
chrisjensenuk: 👍

PR #180, opened 2018-06-27
ER-444 unsuccessful application

chrisjensenuk:
- Wired up navigation - application review route is now: `/accounts/XXXXXX/vacancies/{guid}/applications/{guid}/`
- Added FluentValidation to Employer.Web to do multiple field validation on the `ApplicationReviewEditModel` ViewModel.
- Renamed `ApplicationReviewSuccessfulEvent` to `ApplicationReviewedEvent` and it is published for both successful and unsuccessful commands (Currently we don't need to differentiate between the two for the handlers we currently have)
- Deleted some unused Attribute validation validators


src/Shared/Recruit.Vacancies.Client/Application/Validation/Fluent/ApplicationReviewValidator.cs
```diff
@@ -0,0 +1,37 @@
+﻿using Esfa.Recruit.Vacancies.Client.Domain.Entities;
+using FluentValidation;
+
+namespace Esfa.Recruit.Vacancies.Client.Application.Validation.Fluent
+{
+    public class ApplicationReviewValidator : AbstractValidator<ApplicationReview>
+    {
+        public const int CandidateFeedbackMaxLength = 200;
+
+        public const string OutcomeRequired = "You must select either successful or unsuccessful";
```
sachpatel: As mentioned, I think we should put these on Confluence so the error message content can be reviewed and tracked. Error codes would also be good because of the other systems that will use the client package in future just like Vacancy

src/Shared/UnitTests/Vacancies.Client/Application/VacancyValidation/ApplicationReviewTests.cs
```diff
@@ -0,0 +1,79 @@
...
+using FluentAssertions;
+using Xunit;
+
+namespace Esfa.Recruit.Vacancies.Client.UnitTests.Vacancies.Client.Application.VacancyValidation
+{
+    public class ApplicationReviewTests
+    {
+        public class ShouldRequireCandiateFeedbackIfUnsuccessfulTestData : TheoryData<string, string>
+        {
+            public ShouldRequireCandiateFeedbackIfUnsuccessfulTestData()
```
sachpatel: Candidate*

src/Shared/Recruit.Vacancies.Client/Application/Validation/Fluent/ApplicationReviewValidator.cs
chrisjensenuk: 👍 Will do.  I think we should review the error codes when we need them.

src/Shared/UnitTests/Vacancies.Client/Application/VacancyValidation/ApplicationReviewTests.cs
chrisjensenuk: Gah!

PR #183, opened 2018-07-06
ER-586 - Part one navigation

chrisjensenuk:
I needed to add another bool property to the `Vacancy` entity called `HasCompletedPart1`.  I need this so I can tell if users have seen the Search Preview or not (Step 6 of 6).

So existing vacancies need to have `HasCompletedPart1` set to true if they have completed part one.  I've created an update script for this.

There is now 2 types of routing for Part 1. The step-by-step progression and the there-and-back from the vacancy preview.  To differentiate between the two I've introduce the term `wizard` to describe the step-by-step behaviour.


chrisjensenuk:
Sure. Its Step 6 of 6 that's the problem - the search preview. At the moment we have no way of knowing if it's the users 1st or nth visit to this page so we need a persistent way of knowing whether they've seen this page before so we can change the routing.

PR #185, opened 2018-07-09
ER-593 rename type field in query views collection to view type

sachpatel:


sachpatel:
With some of the changes made in this branch, if we wanted to exclude a change script from being run we would rename the script file to start with an `_` so `documentMIgration.js` ignores it from running.

src/Data/CosmosDb/010-queryViews_LiveVacancy_Insert_DefaultField_ApplicationMethod.js
```diff
@@ -1,55 +1,59 @@
...
-}
-
-do {
-    var matchedDocs = db.queryViews.aggregate([
-        {
-            $match: query
-        },
-        {
-            $sort: { "lastUpdated": 1 }
+{
```
chrisjensenuk: Do we need these braces? Should this be an IIFE? I don't understand where the entry point is, I think I'm missing something.

sachpatel: I wanted to be a ES6 hipster like Ozzy Scott and use `let` and `const`. Of course Mongoshell is not browser based JS or server side the same way as Node is, so unfortunately we can't use ES6 modules or AMD/Require so the `load` from documentMigration.js is like one context and `var` declarations get hoisted into global scope. I was iffy about an *IIFE* because of the way it would get declared and called so I have used `{` to introduce a scope. the documentMigration.js is the sort of entry point of where the scripts will be invoked from.

sachpatel: Not really happy about all this scripting or the complexity of these scripts but this is all we have at the moment. We should move to stored procedures once we have the partition keys sorted.

PR #188, opened 2018-07-13
ER-565 - Vacancies must contain the word 'apprentice' or 'apprenticeships' in the title.

sachpatel:
Note that you can create a vacancy with the title being `apprentice`. This is valid and is what we allow on RAA V1 and the vacancy API.


sachpatel:
@mgwilliam You are not wrong but a simple `string.Contains` with case insensitivity is not enough as we would need to check that `apprentice` isn't part of another word. As it is and it was not mentioned in the story we would not allow `(apprentice)`. Also I chose this as it doesn't involve creating more `if` conditions. We are only validating the title field on the title page and on full vacancy submit so we are not calling validation on this field every time the Vacancy object is being saved.

mgwilliam:
fair enough - just curious as i did it differently in the QA POC work

PR #189, opened 2018-07-13
ER-624 - Ordering of applications viewed on live and closed vacancy summary preview

sachpatel:


sachpatel:
@chrisjensenuk Yeah, because its a display ticket I am okay with this for now. I did think about the ordering and being explicit which is kind of why I cast the enum. Really I want to get this onto test where Sultan has vacancies with multiple applications in different statuses. Will add a tech debt entry on Confluence to address this needing to be more explicit or a test that checks changes maintain the expected order.

PR #193, opened 2018-07-19
ER-586 part one navigation

chrisjensenuk:
**Updated**
- If vacancy has not completed part one then we know we must always display the wizard
- If the vacancy has completed part one but has not started part two then we use the querystring parameter to decide whether to show the wizard or not (as both wizard and non-wizard are valid in this state)
- If the vacancy has started part two then we never show the wizard.

**Original**
- Added `Utility.VacancyHasStartedPartTwo()` to determine whether to resume a vacancy on 'Step 6-6' or the vacancy preview screen.
- Simplified the dashboard status filter logic so the various views that redirect to the dashboard don't need to worry about the cookie.
- Split out Utility Tests into different classes

### DIFF
--- a/src/Employer/Employer.Web/Controllers/LogoutController.cs
+++ b/src/Employer/Employer.Web/Controllers/LogoutController.cs
@@ -5,22 +5,27 @@ using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Hosting;
 
 namespace Esfa.Recruit.Employer.Web.Controllers
 {
     [Route(RoutePrefixPaths.AccountRoutePath)]
     public class LogoutController : Controller
     {
+        private readonly IHostingEnvironment _hostingEnvironment;
         private readonly ExternalLinksConfiguration _externalLinks;
 
-        public LogoutController(IOptions<ExternalLinksConfiguration> externalLinksOptions)
+        public LogoutController(IHostingEnvironment hostingEnvironment, IOptions<ExternalLinksConfiguration> externalLinksOptions)
         {
+            _hostingEnvironment = hostingEnvironment;
             _externalLinks = externalLinksOptions.Value;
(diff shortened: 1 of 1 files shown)


src/Employer/Employer.Web/ViewModels/Part1/Employer/EmployerViewModel.cs
```diff
@@ -17,6 +17,10 @@ public class EmployerViewModel : EmployerEditModel
             nameof(EmployerEditModel.AddressLine4),
             nameof(EmployerEditModel.Postcode)
         };
+
+        public bool IsWizard { get; set; }
```
sachpatel: Could you put these properties in an abstract class that the part 1 viewmodel inherit from

src/Employer/Employer.Web/ViewModels/Part1/Title/TitleViewModel.cs
```diff
@@ -1,4 +1,6 @@
 ﻿using System.Collections.Generic;
+using System.Diagnostics.SymbolStore;
```
sachpatel: ?

src/Employer/Employer.Web/ViewModels/Part1/Employer/EmployerViewModel.cs
chrisjensenuk: Unfortunately no because they already inherit from their respective Edit Models

src/Employer/Employer.Web/ViewModels/Part1/Title/TitleViewModel.cs
chrisjensenuk: Wha?! Will remove

chrisjensenuk:
@sachpatel I'll change the wizard querystring  from a bool to a string and handle unexpected values.

chrisjensenuk:
@sachpatel It should be fine as it is. The DashboardController.ManageDashboardFilter defaults to show all the vacancies if the statusFilter cannot be cast to an enum.

PR #200, opened 2018-07-31
ER-318 QA Statistics

chrisjensenuk:
- I have kept the existing QA dashboard HTML table for now. I've added todo comments to highlight what needs to be removed from the VM, projection, etc.
- Created QA Dashboard projection
- Created `BankHolidayJob` to download bank holiday data from `https://www.gov.uk/bank-holidays.json` every day.
- Created `QaDashboardJob` to update the QA dashboard every 5 minutes.
- Created `QaDashboardService` which is called by the `QaDashboardJob` and `UpdateQaDashboardOnReview` event handler


src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/BankHolidayService.cs
```diff
@@ -0,0 +1,56 @@
...
+    public class BankHolidayService : IBankHolidayService
+    {
+        private readonly IOptions<BankHolidayConfiguration> _config;
+        private readonly IReferenceDataReader _referenceDataReader;
+        private readonly IReferenceDataWriter _referenceDataWriter;
+        private readonly ILogger<BankHolidayService> _logger;
+
+        public BankHolidayService(IOptions<BankHolidayConfiguration> config, IReferenceDataReader referenceDataReader, IReferenceDataWriter referenceDataWriter, ILogger<BankHolidayService> logger)
+        {
+            _config = config;
```
sachpatel: store BankHolidayConfiguration in field instead of the IOptions wrapped one?

PR #201, opened 2018-08-02
ER-494 wage reference data

leewadhams:
Made the reference data provider class more generic.


src/Shared/UnitTests/Vacancies.Client/Application/Services/NationalMinimumWageServiceTests.cs
```diff
@@ -0,0 +1,152 @@
...
+        {
+            _mockLogger = new Mock<ILogger<NationalMinimumWageService>>();
+            var mockDataProvider = new Mock<IMinimumWageDataProvider>();
+            mockDataProvider.Setup(x => x.GetMinimumWageData()).ReturnsAsync(GetTestData());
+
+            _service = new NationalMinimumWageService(mockDataProvider.Object, _mockLogger.Object);
+        }
+
+        [Fact]
+        public void ShouldPickCorrectWageRageBasedOnDate()
```
sachpatel: 😡 Rage

chrisjensenuk: Took me a while to work out why @sachpatel was so upset! ShouldPickCorrectWage**Rage**BasedOnDate

PR #206, opened 2018-08-09
Use Qualifications list from reference data store

leewadhams:


src/Employer/Employer.Web/Mappings/VacancyDisplayViewModelMapper.cs
```diff
@@ -19,22 +19,22 @@ public class DisplayVacancyViewModelMapper
...
                 IOptions<ExternalLinksConfiguration> externalLinksOptions,
                 IEmployerVacancyClient client)
         {
             _mapService = mapService;
             _wageService = wageService;
-            _qualificationsConfiguration = qualificationsConfigOptions.Value;
             _externalLinksConfiguration = externalLinksOptions.Value;
             _client = client;
+
+            _qualifications = _client.GetCandidateQualificationsAsync().Result.QualificationTypes;
```
sachpatel: Could we not use Lazy here and in the ctor for ReviewMapper

leewadhams: Yes we could do. What I meant to do was move it into the mapping method like we do with the programmes. Or we could move both to be lazy initialised. I guess using the lazy initializer makes it easier to re-use if other methods are added to the class.

leewadhams:
@sachpatel mapping classes updated as suggested.

PR #207, opened 2018-08-10
ER-593 changes for query store having view type shard key partition

sachpatel:
I have not been able to test my change script against a Cosmos sharded collection but have tested it against a collection in a local mongo container.


src/Shared/Recruit.Vacancies.Client/Infrastructure/QueryStore/IQueryStore.cs
```diff
@@ -8,12 +8,12 @@ public interface IQueryStore
...
-        Task<T> GetAsync<T>(string key) where T : QueryProjectionBase;
+        Task<T> GetAsync<T>(string typeName, string key) where T : QueryProjectionBase;
 
         Task UpsertAsync<T>(T item) where T : QueryProjectionBase;
 
-        Task<bool> DeleteAsync<T>(string key) where T : QueryProjectionBase;
-
         Task RecreateAsync<T>(IList<T> items) where T : QueryProjectionBase;
+
+        Task<bool> DeleteAsync<T>(string typeName, string key) where T : QueryProjectionBase;
```
mgwilliam: does this need to return?

PR #209, opened 2018-08-23
ER-641 code clean up

leewadhams:
See ticket for details.


src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/IEmployerDashboardService.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace Esfa.Recruit.Vacancies.Client.Infrastructure.Services
 {
-    public interface IDashboardService
+    public interface IEmployerDashboardService
```
sachpatel: Could this be left as `IDashboardService` as the other dashboards could also implement this interface and the method names are general.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Events/QueueNames.cs
```diff
@@ -4,7 +4,7 @@ public static class QueueNames
     {
         public const string VacancyEventsQueueName = "vacancy-events-queue";
         public const string VacancyReviewEventsQueueName = "vacancy-review-events-queue";
-        public const string DashboardQueueName = "dashboard-queue";
+        public const string EmployerDashboardQueueName = "employer-dashboard-queue";
```
sachpatel: Would you mind prefixing this queue name with `generate-`. This was my fault for giving it a bad name in the first place.

src/src.sln
```diff
@@ -0,0 +1,6 @@
+
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 2013
```
sachpatel: 🤣

src/Shared/Recruit.Vacancies.Client/Domain/Events/ApplicationSubmittedEvent.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Esfa.Recruit.Vacancies.Client.Domain.Messaging;
+using MediatR;
+
+namespace Esfa.Recruit.Vacancies.Client.Domain.Events
+{
+    // Note: Doesn't implement IApplicationReviewEvent as it's an externally published event.
```
sachpatel: Should we maybe put this declaration closer to the one job that uses it as it seems very specific maybe.

src/Employer/Employer.Web/Mappings/Extensions/IApprenticeshipProgrammeMapperExtensions.cs
```diff
@@ -3,9 +3,9 @@
 using Esfa.Recruit.Employer.Web.ViewModels.Part1.Training;
 using Esfa.Recruit.Vacancies.Client.Domain.Entities;
 
-namespace Esfa.Recruit.Employer.Web.Extensions
+namespace Esfa.Recruit.Employer.Web.Mappings.Extensions
 {
-    public static class IApprenticeshipProgrammeExensions
+    public static class IApprenticeshipProgrammeMapperExtensions
```
sachpatel: Does ApprenticeshipProgrammes type need an interface?It has resulted in this class sounding interfacey 🦊

src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/IEmployerDashboardService.cs
leewadhams: 👍

src/Shared/Recruit.Vacancies.Client/Infrastructure/Events/QueueNames.cs
leewadhams: Done

src/src.sln
leewadhams: Removed. That was interestingly created by either Visual Studio Code or Rider.

src/Shared/Recruit.Vacancies.Client/Domain/Events/ApplicationSubmittedEvent.cs
leewadhams: But it is still a domain event isn't it? We're just acting on it as opposed to raising it.

sachpatel: 👍 I was just thinking because it is used in one place.

PR #212, opened 2018-08-24
Feature/ER-635 vacancy search

dev4valtech:
Search is using query string to the same get endpoint as of dashboard.


src/QA/QA.Web/Controllers/DashboardController.cs
```diff
@@ -17,9 +20,26 @@ public DashboardController(DashboardOrchestrator orchestrator)
         [HttpGet("/", Name = RouteNames.Dashboard_Index_Get)]
         public async Task<IActionResult> Index()
         {
-            var vm = await _orchestrator.GetDashboardViewModelAsync();
+            var vm = await _orchestrator.GetDashboardViewModelAsync().ConfigureAwait(false);
```
chrisjensenuk: I don't think we need to use ConfigureAwait in Core. https://stackoverflow.com/questions/13489065/best-practice-to-call-configureawait-for-all-server-side-code

src/QA/QA.Web/Controllers/DashboardController.cs
```diff
@@ -17,9 +20,26 @@ public DashboardController(DashboardOrchestrator orchestrator)
...
+
+        /// <summary>
+        /// Opted to use POST instead of GET as it keeps the URL clean of query string and path
+        /// </summary>
+        [HttpPost("/", Name = RouteNames.Dashboard_Index_Post)]
+        public async Task<IActionResult> Index(string searchTerm)
+        {
+            var vm = await _orchestrator.GetDashboardViewModelAsync().ConfigureAwait(false);
+
+            //I would prefer not to go forward if search term is empty
```
chrisjensenuk: Don't think we need this comment.  I think its OK to keep the controller dumb and let client handle empty queries.  That's the approach we've taken for validation, etc so fits into this approach.

src/QA/QA.Web/Views/Dashboard/Index.cshtml
```diff
@@ -1,59 +1,97 @@
...
             </tbody>
         </table>
     </div>
 </div>
+<br /><br />
+
+<h2 asp-condition="@Model.IsPostBack" class="heading-medium no-bottom-margin">Search results for [@Model.LastSearchTerm]</h2>
 
-<br/><br/>
+<span asp-condition="@Model.IsPostBack && Model.SearchResults.Any() == false">No results</span>
```
chrisjensenuk: We're trying to keep any display logic in the VMs and out of the Views.  So maybe create a `DisplayNoResultsMessage` bool in the VM instead.

src/QA/QA.Web/ViewModels/DashboardViewModel.cs
```diff
@@ -9,5 +9,12 @@ public class DashboardViewModel
         public int TotalVacanciesResubmitted { get; set; }
 
         public List<ReviewDashboardItem> AllReviews { get; set; }
+
+        public string SearchTerm { get; set; }
+
+        public string LastSearchTerm { get; set; }
+
+        public List<VacancyReviewSearchModel> SearchResults { get; set; } = new List<VacancyReviewSearchModel>();
+        public bool IsPostBack { get; set; }
```
chrisjensenuk: I don't think the view should care about postbacks.  We tend to create bool properties and helper functions in the VM to change the view.  Maybe just create a `DisplayNoResultsMessage` property in the VM instead (I think this is only dependency on knowledge of a postback)

src/QA/QA.Web/Orchestrators/DashboardOrchestrator.cs
```diff
@@ -20,19 +22,51 @@ public DashboardOrchestrator(IQaVacancyClient vacancyClient)
...
+                VacancyTitle = vacancyReviewSearch.Title,
+                ReviewId = vacancyReviewSearch.Id,
+                SubmittedDate = vacancyReviewSearch.SubmittedDate.ToLocalTime()
+            };
+        }
+
+        private static string GetElapsedTime(DateTime? value)
+        {
+            if (value == null) return string.Empty;
+            var diff = DateTime.Now - value.Value;
```
chrisjensenuk: Use _timeProvider instead of DateTime.Now

src/QA/QA.Web/Views/Dashboard/Index.cshtml
```diff
@@ -1,59 +1,97 @@
...
 
+<div asp-condition="@Model.SearchResults.Any()" class="grid-row">
```
chrisjensenuk: Maybe `@Model.DisplaySearchResults`

src/QA/QA.Web/Views/Dashboard/Index.cshtml
```diff
@@ -1,59 +1,97 @@
...
+        {
+            <tr>
+                <td>@vacancy.EmployerName</td>
+                <td>@vacancy.VacancyTitle</td>
+                <td>@vacancy.VacancyReference</td>
+                <td>@vacancy.SubmittedDate.AsGdsDateTime()</td>
+                <td>@vacancy.ClosingDate.AsGdsDate()</td>
+                <td>
+                    <a asp-condition="@vacancy.IsAvailableForReview" asp-route="@RouteNames.Vacancy_Review_Get" asp-route-reviewId="@vacancy.ReviewId">Review</a>
+                    <span asp-condition="@vacancy.IsAvailableForReview == false">@vacancy.AssignmentInfo</span>
```
chrisjensenuk: We've been adding inverses into VMs for this. E.g. `@Model.IsNotAvailableForReview`

src/Shared/Recruit.Vacancies.Client/Infrastructure/Client/QaVacancyClient.cs
```diff
@@ -72,14 +74,41 @@ public Task<Qualifications> GetCandidateQualificationsAsync()
...
+        public async Task<List<VacancyReviewSearch>> GetSearchResultsAsync(string searchTerm)
+        {
+            if (TryGetVacancyReference(searchTerm, out var vacancyReference))
+            {
+                return await _reviewRepository.SearchAsync(vacancyReference).ConfigureAwait(false);
+            }
+            return new List<VacancyReviewSearch>();
+        }
+
+        /// <summary>
```
chrisjensenuk: I'd kill this comment

src/Shared/Recruit.Vacancies.Client/Infrastructure/Client/QaVacancyClient.cs
```diff
@@ -72,14 +74,41 @@ public Task<Qualifications> GetCandidateQualificationsAsync()
...
+        /// This could be somewhere more common, we can refactor when we come across a second use
```
sachpatel: This could be a extension method on the `string` class. Best place to put it would be in the shared project so both Employer and QA app can use it. Also this shouldn't be called from the client but by the web projects since the `VAC` prefix is purely a display/front-end concern.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Client/QaVacancyClient.cs
```diff
@@ -72,14 +74,41 @@ public Task<Qualifications> GetCandidateQualificationsAsync()
...
+        /// </summary>
+        private static bool TryGetVacancyReference(string value, out long vacancyReference)
+        {
+            vacancyReference = 0;
+            if (string.IsNullOrEmpty(value)) return false;
+            
+            var regex = new Regex(@"^(VAC)?(\d{10})$", RegexOptions.IgnoreCase);
```
sachpatel: maybe just check `value.Replace("vac", string.Empty).Replace("VAC", string.empty)` and then long try parse the result. More to avoid Regexing

src/Shared/Recruit.Vacancies.Client/Infrastructure/Repositories/MongoDbVacancyReviewRepository.cs
```diff
@@ -22,6 +22,37 @@ public MongoDbVacancyReviewRepository(ILogger<MongoDbVacancyReviewRepository> lo
...
+            var filter = (filterBuilder.Eq(r => r.Status, ReviewStatus.PendingReview)
+                         | filterBuilder.Eq(r => r.Status, ReviewStatus.UnderReview)) 
+                         & filterBuilder.Eq(r => r.VacancyReference, vacancyReference);
+            var collection = GetCollection<VacancyReview>();
+            
+            var result = await RetryPolicy
+                .ExecuteAsync( 
+                    context => collection
+                        .Find(filter)
+                        .Project(r => new VacancyReviewSearch()
```
sachpatel: This ,ethod should return the VacancyReview and then that will get mapped to the VacancyReviewSearch from the consuming code.

sachpatel:
I think we can do the search as part of the main dashboard `index` http get endpoint. Where you can have some conditional logic to perform search if the querystring `searchFilter` is provided or not and that way we can keep the filter in the url.

chrisjensenuk:
I think I'd prefer to see the search as a GET too.

src/QA/QA.Web/ViewModels/DashboardViewModel.cs
dev4valtech: I think this needs wider consultation. The page is designed in unconventional way. We perhaps need to separate results page, that way we can use query string while doing a GET request. It doesn't make sense to fit it all on dashboard.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Client/QaVacancyClient.cs
dev4valtech: I think regex works better in this case as not only is checks for the string case insensitively, it also makes sure that the number has correct length.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Client/QaVacancyClient.cs
dev4valtech: the search term is going to be common across multiple fields, so the code is appropriately placed on the client.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Repositories/MongoDbVacancyReviewRepository.cs
dev4valtech: With VacancySnapshot added, Vacancy Review is now a bulky document, hence I am projecting to retrieve only required fields.

PR #217, opened 2018-08-31
ER-426 ability to extend live vacancy dates

sachpatel:
I have used the query string to pass state between proposed date changes, this is because I wanted something temporary but didn't want to use cookies or temp data. Temp data is not ideal where multiple requests could be made by the same user.

I have used the `FromQuery` attribute on some action parameters because I was not able to use them on a model class properties. Another option is to create a custom model binder but didn't think that would be appropriate for this small case.

Any changes to vacancy dates being previewed will be loaded again on the edit dates page where possible (from query string) so the user is changing from the dates they expect to see.

One thing that has not been handled well for the user experience is, if the dates in the query string portion of ManageVacancy url is tampered with and includes an invalid date, on submission, the user will be redirected to the EditDates page but will not have a validation error. Happy to take any suggestions on what to display.


src/Employer/Employer.Web/Orchestrators/VacancyViewOrchestrator.cs
```diff
@@ -0,0 +1,160 @@
+﻿using Esfa.Recruit.Employer.Web.Configuration;
+using Esfa.Recruit.Employer.Web.Mappings;
+using Esfa.Recruit.Employer.Web.Models;
+using Esfa.Recruit.Employer.Web.RouteModel;
+using Esfa.Recruit.Employer.Web.ViewModels;
+using Esfa.Recruit.Employer.Web.ViewModels.VacancyManage;
+using Esfa.Recruit.Vacancies.Client.Domain.Entities;
+using Esfa.Recruit.Shared.Web.Extensions;
```
leewadhams: Redundant

src/Employer/Employer.Web/Controllers/VacancyManageController.cs
```diff
@@ -29,36 +34,53 @@ public async Task<IActionResult> DisplayVacancy(VacancyRouteModel vrm)
...
+            if (vacancy.Status != VacancyStatus.Live)
+            {
+                return RedirectToRoute(RouteNames.DisplayVacancy_Get);
+            }
 
-        [HttpGet("view", Name = RouteNames.DisplayFullVacancy_Get)]
-        public async Task<IActionResult> DisplayFullVacancy(VacancyRouteModel vrm)
-        {
-            var vacancy = await _orchestrator.GetVacancy(vrm);
+            DateTime parsedClosingDate = DateTime.MinValue;
```
leewadhams: Is there a more succinct/easier to read way of writing this?

https://github.com/SkillsFundingAgency/das-recruit/blob/7362f28c5cf1e3d460f751fb099107b2d372a0db/src/Employer/Employer.Web/Controllers/VacancyManageController.cs#L42-L58

leewadhams:
@sachpatel what was the reason for not using the `FromQuery` in a model class?

chrisjensenuk:
Hey Sachin. Whats the reasoning behind not using a cookie? If we put this in a cookie then we won't have to make any changes to the routing and do all the querystring marshalling and custom date formats. etc.

src/Employer/Employer.Web/Orchestrators/EditVacancyDatesOrchestrator.cs
```diff
@@ -0,0 +1,152 @@
...
+                        resp.Data.StartMonth = parsedStartDate.Month.ToString("00");
+                        resp.Data.StartYear = parsedStartDate.Year.ToString();
+                    }
+                }
+            }
+        }
+
+        public async Task<EditVacancyDatesViewModel> GetEditVacancyDatesViewModel(EditVacancyDatesEditModel m)
+        {
+            var resp = await GetEditVacancyDatesViewModel((VacancyRouteModel)m, m.ClosingDate.ToDateQueryString(), m.StartDate.ToDateQueryString());
```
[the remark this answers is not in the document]
sachpatel: will look into

src/Employer/Employer.Web/Controllers/VacancyManageController.cs
sachpatel: I struggled to think of a way to to make this look more readable, partly because of the return redirects. If you have any suggestion?

src/Employer/Employer.Web/Controllers/EditVacancyDatesController.cs
```diff
@@ -0,0 +1,101 @@
...
+    {
+        private readonly EditVacancyDatesOrchestrator _orchestrator;
+
+        public EditVacancyDatesController(EditVacancyDatesOrchestrator orchestrator)
+        {
+            _orchestrator = orchestrator;
+        }
+
+        [HttpGet("edit-dates", Name = RouteNames.VacancyEditDates_Get)]
+        public async Task<IActionResult> EditVacancyDates(VacancyRouteModel vrm, [FromQuery]string proposedClosingDate, [FromQuery]string proposedStartDate)
```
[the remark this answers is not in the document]
sachpatel: if they are changing the dates they have already changed the review of.

src/Employer/Employer.Web/Controllers/VacancyManageController.cs
```diff
@@ -29,36 +34,53 @@ public async Task<IActionResult> DisplayVacancy(VacancyRouteModel vrm)
...
+            TempData.Add(TempDataKeys.DashboardInfoMessage, string.Format(InfoMessages.VacancyUpdated, vacancy.Title));
+            return RedirectToRoute(RouteNames.Dashboard_Index_Get);
         }
 
         private IActionResult HandleRedirectOfDraftVacancy(Vacancy vacancy)
         {
             if (Utility.VacancyHasCompletedPartOne(vacancy))
             {
-                if (Utility.VacancyHasStartedPartTwo(vacancy) == false)
-                    return RedirectToRoute(RouteNames.SearchResultPreview_Get);
```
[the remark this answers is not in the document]
sachpatel: Did not notice this. will add back in.

sachpatel:
@chrisjensenuk Did not use cookie as I thought this is something for the moment that is more temporary for state passing between requests.

src/Employer/Employer.Web/Controllers/EditVacancyDatesController.cs
```diff
@@ -0,0 +1,101 @@
...
+using System.Threading.Tasks;
+using Esfa.Recruit.Employer.Web.Configuration.Routing;
+using Esfa.Recruit.Employer.Web.Orchestrators;
+using Esfa.Recruit.Employer.Web.RouteModel;
+using Esfa.Recruit.Employer.Web.ViewModels.EditVacancyDates;
+using Esfa.Recruit.Employer.Web.ViewModels.VacancyManage;
+using Esfa.Recruit.Vacancies.Client.Domain.Entities;
+using Microsoft.AspNetCore.Mvc;
+using Esfa.Recruit.Shared.Web.Extensions;
+using Esfa.Recruit.Employer.Web.QueryStringModels;
```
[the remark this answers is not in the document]
sachpatel: Will remove

src/Employer/Employer.Web/Orchestrators/VacancyViewOrchestrator.cs
sachpatel: will remove

sachpatel:
For security reasons (Should not redisplay data from a query string back on a page), will make changes to this PR to use a single vacancy specific cookie for storing the previewable change in vacancy dates as opposed to state being passed on the query string.

sachpatel:
Have changed to use a cookie to store the dates being changed to. The cookie has a life time of 5 minutes and gets cleared on successful publish or cancel from the preview or edit dates pages.

src/Employer/Employer.Web/Configuration/QueryString.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+
+namespace Esfa.Recruit.Employer.Web.Configuration
+{
+    public static class QueryString
```
chrisjensenuk: Do we still need this class?

src/Employer/Employer.Web/Configuration/EsfaCookieOptions.cs
```diff
@@ -11,5 +12,13 @@ public static class EsfaCookieOptions
             SameSite = SameSiteMode.Strict,
             HttpOnly = true
         };
+
+        public static CookieOptions GetShortLifeHttpCookieOption(IHostingEnvironment env) => new CookieOptions
+        {
+            Secure = !env.IsDevelopment(),
+            SameSite = SameSiteMode.Strict,
+            HttpOnly = true,
+            Expires = DateTimeOffset.UtcNow.AddMinutes(5)
```
dev4valtech: A user session lasts for longer than 5 minutes, it is likely that the cookies will be gone if the user is distracted for over 5 minutes, have you accounted for if the cookies are not found or considered leaving the cookie for longer duration?

src/Employer/Employer.Web/Configuration/QueryString.cs
dev4valtech: Is this class needed?

src/Employer/Employer.Web/Controllers/EditVacancyDatesController.cs
```diff
@@ -0,0 +1,118 @@
...
+                response.AddErrorsToModelState(ModelState);
+            }
+
+            if (!ModelState.IsValid)
+            {
+                var vm = await _orchestrator.GetEditVacancyDatesViewModel(m);
+                return View(vm);
+            }
+
+            Response.Cookies.Append(string.Format(CookieNames.VacancyProposedClosingDate, m.VacancyId), DateTime.Parse(m.ClosingDate).ToShortDateString(), EsfaCookieOptions.GetShortLifeHttpCookieOption(_hostingEnvironment));
```
chrisjensenuk: Is it possible to have just one cookie and serialize a DTO that contains the dates as DateTime properties? Maybe even store the vacancyId in the DTO so we can be sure the value pertain to the right vacancy. In future we could add further proposed changes to this cookie.

src/Employer/Employer.Web/Configuration/QueryString.cs
sachpatel: oops, will remove

sachpatel: have removed in next commit

src/Employer/Employer.Web/Configuration/EsfaCookieOptions.cs
sachpatel: I put 5 minutes, which I will speak to PO/BA about tomorrow. The reason I put 5 is because I do not know how long is left of their auth timeout. If the cookies are not found they will see old values of the actual vacancy and cannot submit changes as there are none. This will not block the user form making changes.

src/Employer/Employer.Web/Controllers/EditVacancyDatesController.cs
sachpatel: Would prefer to avoid serializing and deserailizing json for the moment.

src/Employer/Employer.Web/Configuration/EsfaCookieOptions.cs
dev4valtech: Make cookie valid only for that session, so it expires if the user session expires?

PR #218, opened 2018-09-03
ER-661 store geo code strategy method with vacancy

sachpatel:


mgwilliam:
@sachpatel is setting the method to unspecified for existing data correct?

PR #222, opened 2018-09-04
ER-453 Approve vacancy

chrisjensenuk:
A big one I'm afraid.

- I've added `vacancyReviewsApprovedCount` & `vacancyReviewsApprovedFirstTimeCount` to the user entity. These 'user statistics' get recalculated on every vacancy review approval.
- I've added Fluent validation to the QA website
- Review page is built from review snapshot data and not the vacancy entity.
- Added new provider`NextVacancyReviewProvider` which returns the 'next' vacancy for review for the specified user.
- Added a TagHelper `EsfaReviewCheckboxTagHelper` to help build the checkboxes on the Review screen.
- Added ViewModel validation for the Reviewer comment and validation in the client for Review validation.


chrisjensenuk:
Will do

src/QA/QA.Web/ViewModels/Validations/ReviewEditModelValidator.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Esfa.Recruit.Vacancies.Client.Application.Validation.Fluent;
+using FluentValidation;
+
+namespace Esfa.Recruit.Qa.Web.ViewModels.Validations
+{
+    public class ReviewEditModelValidator : AbstractValidator<ReviewEditModel>
```
leewadhams: Is this a move away from the pattern we use in Employer?

src/Shared/Recruit.Vacancies.Client/Infrastructure/Providers/NextVacancyReviewProvider.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System.Linq;
+using System.Threading.Tasks;
+using Esfa.Recruit.Vacancies.Client.Application.Providers;
+using Esfa.Recruit.Vacancies.Client.Domain.Entities;
+using Esfa.Recruit.Vacancies.Client.Domain.Repositories;
+using Esfa.Recruit.Vacancies.Client.Domain.Services;
+
+namespace Esfa.Recruit.Vacancies.Client.Infrastructure.Providers
+{
+    public class NextVacancyReviewProvider : INextVacancyReviewProvider
```
leewadhams: This sounds more like an Application Service i.e. working out what the next vacancy to look at is.

src/QA/QA.Web/ViewModels/Validations/ReviewEditModelValidator.cs
chrisjensenuk: I can see the benefit of pushing Vacancy validations through to the client as we modify it over a series of pages. However for simpler entity interactions I think there is no benefit in adding this extra complexity (E.g. Overhead of marshalling entity validations to their VM counterparts).
So for VacancyReview I've thought adding view model validations and equivalent entity validations is the simpler solution for this ticket. Happy to change if needed.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Providers/NextVacancyReviewProvider.cs
chrisjensenuk: Yeah. I flip-flopped on the naming of this.  As it returns a reviewId and doesn't change anything I landed on Provider. Happy to change. 👍

leewadhams: I think it should be going by it definition. Means it will move from Infrastructure to Application.

src/QA/QA.Web/ViewModels/Validations/ReviewEditModelValidator.cs
leewadhams: Yep I take your point. We will just have the downside of having to keep the 2 validators in sync.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Repositories/MongoDbVacancyReviewRepository.cs
```diff
@@ -102,5 +114,36 @@ public Task<List<VacancyReview>> GetForVacancyAsync(long vacancyReference)
                 .Find(filter)
                 .ToListAsync(), new Context(nameof(GetForVacancyAsync)));
         }
+
+        public async Task<int> GetApprovedFirstTimeCountAsync(string submittedByUserId)
```
dev4valtech: This is the subset of the Approved query, perhaps consider getting a list and doing the count from in memory list instead of two different queries.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Repositories/MongoDbVacancyReviewRepository.cs
```diff
@@ -102,5 +114,36 @@ public Task<List<VacancyReview>> GetForVacancyAsync(long vacancyReference)
...
+        {
+            var filterBuilder = Builders<VacancyReview>.Filter;
+            var filter = filterBuilder.Eq(r => r.SubmittedByUser.UserId, submittedByUserId) &
+                         filterBuilder.Eq(r => r.Status, ReviewStatus.Closed) &
```
dev4valtech: Is this criteria necessary after checking for manual outcome?

src/Shared/Recruit.Vacancies.Client/Infrastructure/Repositories/MongoDbVacancyReviewRepository.cs
chrisjensenuk: I don't have enough info on performance to know which will be quicker. My thinking is returning a count is quicker than a projection so that's what I've gone for.  Suggest we just monitor the page load times as the data grows.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Repositories/MongoDbVacancyReviewRepository.cs
chrisjensenuk: Belt and braces. It felt right to include the actual status of the review as part of this query.  If a status was to be reverted (through a data fix for example) then it won't be included in this count.

mgwilliam:
@chrisjensenuk just a couple:

* the `NextVacancyReviewServices` has an extra trailing `s`
* the above service is called from the client... should be an AssignReview command to wrap the logic - see https://skillsfundingagency.atlassian.net/wiki/spaces/RAAV2/pages/283213874/Vacancy+QA+web+application#VacancyQAwebapplication-Command:Assignreview

chrisjensenuk:
I overlooked that reviews can also be assigned via links from search results.  So I've renamed `AssignNextVacancyCommand` to `AssignVacancyCommand`. It now takes an optional ReviewId.  If ReivewId is specified then it will attempt to assign that review to the user.  If it is not specified then it work out what the next review is and then assign that review.

chrisjensenuk:
@dev4valtech has reviewed.

PR #224, opened 2018-09-05
Make WageType nullable

leewadhams:


leewadhams:
Yep. Validation was already in place.

PR #227, opened 2018-09-07
ER-667 custom skills added as checkboxes

leewadhams:
There may well be a much simpler/better way to do this.

To retain ordering of the custom skills entered before they are saved I've introduced an order number for them. It surfaces as a prefix of the `value` on the checkbox `e.g. 1-Programming Ability`


leewadhams:
Oh it was lots of fun! That issue should be fixed.

PR #228, opened 2018-09-10
Feature/ER-629 reviews in progress

dev4valtech:
NextVacancyReviewService is defined in the application layer. This service exposes method GetExpiredAssignationDateTime which was required for filtering the records. However, since the service is in the application layer, this value is read from the Client and passed into the repository.


mgwilliam:
@dev4valtech are ops aware of the new config values?

dev4valtech:
@mgwilliam  Not sure if Lee has communicated, but I am assuming not, I will notify them.

src/QA/QA.Web/Orchestrators/DashboardOrchestrator.cs
```diff
@@ -14,52 +17,71 @@ public class DashboardOrchestrator
...
+                VacancyTitle = qaVacancySummary.Title,
+                ReviewId = qaVacancySummary.Id,
+                SubmittedDate = qaVacancySummary.SubmittedDate.ToLocalTime(),
+                IsAvailableForReview = qaVacancySummary.ReviewAssignedToUserId == vacancyUser.UserId || isAvailableForReview,
+                IsNotAvailableForReview = !isAvailableForReview
             };
         }
 
-        private string GetElapsedTime(DateTime? value)
+        private string GetAssignmentInfo(DateTime? value, string userName)
```
sachpatel: This method is now doing two things and is now more related to the view. Could we not leave the sentences in the view. Only the swappable parts could be worked out here or in the viewmodel. If the razor page isn't the place to put it, I would suggest a static utility method that is called from the razor page.

dev4valtech: I have consciously made the change to calculate the whole field in the orchestrator so I don't have to expose fields like username and timeelapsed which could now have value of "you" and "now" respectively.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Repositories/MongoDbVacancyReviewRepository.cs
```diff
@@ -14,47 +16,71 @@ namespace Esfa.Recruit.Vacancies.Client.Infrastructure.Repositories
...
+        public MongoDbVacancyReviewRepository(
+            ILogger<MongoDbVacancyReviewRepository> logger, IOptions<MongoDbConnectionDetails> details, 
+            ITimeProvider timeProvider) 
             : base(logger, Database, Collection, details)
         {
+            _timeProvider = timeProvider;
         }
 
-        public async Task<List<VacancyReviewSearch>> SearchAsync(long vacancyReference)
+        public Task<List<QaVacancySummary>> SearchAsync(long vacancyReference)
```
leewadhams: Should this summary class be generated in the orchestrator instead of the repo?

dev4valtech: I used projection instead of retrieving the whole document which also has vacancy snapshot now.

PR #233, opened 2018-09-12
ER-629 reviews in progress2

sachpatel:


sachpatel:
@chrisjensenuk Have changed for `AssignedTo`, the time elapsed property will take care of it itself if it is null.

dev4valtech:
ReviewStartedOn will be set to null, as part of manually Unassign story

PR #238, opened 2018-09-17
ER-681 Remove unused CommandId

dev4valtech:


src/Shared/Recruit.Vacancies.Client/Domain/Messaging/CommandBase.cs
```diff
@@ -1,14 +1,6 @@
-﻿using System;
-
-namespace Esfa.Recruit.Vacancies.Client.Domain.Messaging
+﻿namespace Esfa.Recruit.Vacancies.Client.Domain.Messaging
 {
     public abstract class CommandBase
```
leewadhams: Do we need CommandBase class any more?

PR #244, opened 2018-09-19
ER-681 Added CandidateSkillsProvider

dev4valtech:
and refactored the orchestrator to use skill list instead of CandidateSkills


src/Shared/Recruit.Vacancies.Client/Infrastructure/ReferenceData/Skills/CandidateSkillsProvider.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            _referenceDataReader = referenceDataReader;
+            _logger = logger;
+        }
+        public async Task<List<string>> GetCandidateSkillsAsync()
+        {
+            var result = await _referenceDataReader.GetReferenceData<CandidateSkills>();
+            if (result != null)
+                return result.Skills;
+            _logger.LogWarning("Unable to retrieve reference data for qualifications");
```
sachpatel: qualifications or skills? :suspect:

dev4valtech: Fixed

PR #246, opened 2018-09-20
ER-693 - Loading banned phrases and inoffensive profanities

sachpatel:
…data into Cosmos referenceData collection.


mgwilliam:
@sachpatel with the manual loading of reference data does this introduce a risk of wiping out the real reference data in subsequent releases?

sachpatel:
@mgwilliam no.

PR #247, opened 2018-09-20
Fe bug fixes

sdholder:
Three front end fixes ...

Wrapping email address on QA review
Character counts incorrect when carriage return used
Radio button label not selecting input


src/QA/QA.Web/Views/Review/Review.cshtml
```diff
@@ -6,31 +6,49 @@
 }
 <div class="review-container">
     <form asp-route="@postRoute">
+
         <div class="grid-row">
-            <div asp-validation-summary="All" class="error-summary" style="display:@(ViewData.ModelState.IsValid ? "none" : "block")"></div>
+            <div class="column-full">
```
sachpatel: can you put an `asp-condition="@ViewData.ModelState.IsValid" attribute on this div so we don't need the `style` attribute

sdholder: Done

PR #248, opened 2018-09-20
Bugfix ER-704 mongo driver exception adding to retry policy

sachpatel:


src/Shared/Recruit.Vacancies.Client/Infrastructure/Mongo/MongoDbCollectionBase.cs
```diff
@@ -47,6 +47,7 @@ private Polly.Retry.RetryPolicy GetRetryPolicy()
         {
             return Policy
                     .Handle<MongoCommandException>()
+                    .Or<MongoConnectionException>()
```
leewadhams: Do we just make it MongoException to cover anything that inherits from it?

sachpatel: didn't notice that 🤦‍♂️  but yes, will change

PR #253, opened 2018-09-24
ER-601 Upgrade .Net Core version

leewadhams:
- Updated projects to target .net core 2.1
- Fixed compiler warnings introduced by the update.
- Update startup classes with new features/format.
- Replaced Html.Partial with <partial> tag helper.


leewadhams:
There have been a couple of changes around https redirection.

PR #256, opened 2018-09-26
ER-614 employer dashboard

chrisjensenuk:
This is essentially simplifying the dashboard.  I've removed the lotto numbers, filtering functionality and its cookie.


src/Employer/Employer.Web/Controllers/DashboardController.cs
```diff
@@ -34,25 +27,7 @@ public async Task<IActionResult> Dashboard([FromRoute]string employerAccountId,
...
-            ManageDashboardFilter(statusFilter, vm);
-
             return View(vm);
         }
-
-        private void ManageDashboardFilter(string statusFilter, DashboardViewModel vm)
-        {
-            if (!string.IsNullOrEmpty(statusFilter) && Enum.TryParse<VacancyStatus>(statusFilter, out var filter))
-            {
-                vm.Filter = filter;
```
sachpatel: If you are removing this cookie could you please speak with Faizal/Raina/MA about having the cookie removed from their cookies and policies page please?

src/Employer/Employer.Web/Controllers/VacancyViewController.cs
```diff
@@ -20,7 +20,7 @@ public VacancyViewController(VacancyViewOrchestrator orchestrator)
         }
 
         [HttpGet("", Name = RouteNames.DisplayVacancy_Get)]
-        public async Task<IActionResult> DisplayVacancy(VacancyRouteModel vrm)
+        public async Task<IActionResult> DisplayVacancy(VacancyRouteModel vrm, [FromQuery] string showApplications)
```
sachpatel: Unsure about this faux boolean query string parameter. Might need to speak with @mgwilliam about this one.

src/Employer/Employer.Web/Extensions/VacancySummaryExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+                case VacancyStatus.Draft:
+                    return "Edit and submit";
+            }
+
+            return "View";
+        }
+
+        public static string GetLinkRoute(this VacancySummary vacancySummary)
+        {
+            if (vacancySummary.Status == VacancyStatus.Live || vacancySummary.Status == VacancyStatus.Closed)
```
sachpatel: Is the Closed because of the edit dates vacancy because we cannot edit or perform any actions on Closed vacancies yet. I think maybe checking the status is live is all that is needed.

src/Employer/Employer.Web/Controllers/VacancyViewController.cs
chrisjensenuk: @mgwilliam What's your opinion on this? ta

src/Employer/Employer.Web/Extensions/VacancySummaryExtensions.cs
chrisjensenuk: 👍 Manage is just redirecting to View for non-live vacancies so this makes sense

src/Employer/Employer.Web/Controllers/VacancyViewController.cs
mgwilliam: @chrisjensenuk i don't have a preference other than we should be consistent with any other similar instances in the code

dev4valtech: I had a conversation with Chris on this, the boolean is used to decide between two views on the same end point, perhaps we could just have two end points instead.

chrisjensenuk: We have precedent for using pseudo booleans as request parameters.  I think adding another 'view vacancy' endpoint just complicates things.  I'm happy as is it.

mgwilliam: @chrisjensenuk @dev4valtech bear in mind that the dashboard isn't having tabs on for public beta as sorcha has removed them

src/Employer/Employer.Web/Controllers/DashboardController.cs
chrisjensenuk: Faizal on the case

PR #264, opened 2018-10-01
ER-651 Automated QA results in QA Web

chrisjensenuk:
- Created a `TemplateRunner` which can be shared between Employer and QA web.  This class converts the Automated QA 'data' to a human readable message.  Further templates for other rules need to be registered in IoC and the runner will use them.

- Created `FieldIdResolver.cs` (which is a wrapper around GetQualifiedFieldId) which resolves Vacancy properties to a string when you don't have an instance of Vacancy.

- Added a little fix to the profanity check so Qualifications are correctly inspected.


src/Shared/Recruit.Shared.Web/RuleTemplates/RuleTemplateRunner.cs
```diff
@@ -0,0 +1,30 @@
...
+        private readonly IServiceProvider _provider;
+
+        public RuleTemplateRunner(IServiceProvider provider)
+        {
+            _provider = provider;
+        }
+
+        public string ToText(object data, string fieldName)
+        {
+            if (data == null)
```
dev4valtech: Where data is null, may be there is a static message.

src/QA/QA.Web/Mappings/ReviewMapper.cs
```diff
@@ -4,12 +4,16 @@
...
+using Esfa.Recruit.Shared.Web.Mappers;
+using Esfa.Recruit.Shared.Web.RuleTemplates;
 using Esfa.Recruit.Shared.Web.Services;
 using Esfa.Recruit.Vacancies.Client.Application.Providers;
+using Esfa.Recruit.Vacancies.Client.Application.Services;
 using Esfa.Recruit.Vacancies.Client.Domain.Entities;
 using Esfa.Recruit.Vacancies.Client.Domain.Extensions;
 using Esfa.Recruit.Vacancies.Client.Infrastructure.Client;
 using Humanizer;
+using Microsoft.EntityFrameworkCore.Internal;
```
sachpatel: 🔥 Please remove

chrisjensenuk: 👍

src/Shared/Recruit.Shared.Web/RuleTemplates/RuleTemplateRunner.cs
chrisjensenuk: I think returning an empty string is OK here.  I'll filter out results with an empty message instead when building the VM (This should never happen btw).

src/QA/QA.Web/Views/Review/Review.cshtml
```diff
@@ -395,12 +422,28 @@
...
+            var $comments = $("#automatedqarules-selected-comments");
+            $comments.empty();
+            
+            $(".automatedqaresults-checkbox:checked").each(function () {
+                var $automatedQaText = $("<li></li>").text($(this).data("text"));
+                $comments.append($automatedQaText);
+            });
+        }
+
+        toggleSubmitButton();
```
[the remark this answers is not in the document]
chrisjensenuk: We could set the submit button text server side but the JS onload will immediately change it anyway so I don't see the benefit of this extra complication.

`hookupHistory()` is shared by `Review.cshtml` and `ReadonlyReview.cshtml`. The js here is only used on this page. Whether it should go into its own module or not I don't have an opinion on.

chrisjensenuk:
- The Rule Message Template is now determined by RuleId and not the Type of data
- Added a RuleId enum
- Renamed RuleTemplates to RuleMessageTemplates

src/QA/QA.Web/Mappings/ReviewMapper.cs
```diff
@@ -27,13 +27,13 @@ public class ReviewMapper
         private readonly IGeocodeImageService _mapService;
         private readonly IMinimumWageProvider _wageProvider;
         private readonly Lazy<IList<string>> _qualifications;
-        private readonly IRuleTemplateRunner _ruleTemplateRunner;
+        private readonly IRuleTemplateMessageRunner _ruleTemplateRunner;
```
dev4valtech: IRuleMessageTemplateRunner may be, matches MessageTemplate naming convention used else where

src/Shared/Recruit.Shared.Web/RuleTemplates/RuleTemplateMessageRunner.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using Esfa.Recruit.Vacancies.Client.Application.Rules;
+
+namespace Esfa.Recruit.Shared.Web.RuleTemplates
+{
+    public class RuleTemplateMessageRunner : IRuleTemplateMessageRunner
```
dev4valtech: RuleMessageTemplateRunner

PR #268, opened 2018-10-03
ER-651 - Old reviews will have a null AutomatedQaOutcomeIndicators list

chrisjensenuk:


src/QA/QA.Web/Orchestrators/ReviewOrchestrator.cs
```diff
@@ -29,12 +29,15 @@ public ReviewOrchestrator(IQaVacancyClient vacancyClient, ReviewMapper mapper)
 
             var manualQaFieldIndicators = _mapper.GetManualQaFieldIndicators(m);
 
-            foreach (var automatedQaOutcomeIndicator in review.AutomatedQaOutcomeIndicators)
+            if (review.AutomatedQaOutcomeIndicators != null)
```
dev4valtech: Isn't it better to initialize this property with empty collection and check for Any() here instead. An empty list is saved when there are no findings from the rules.

chrisjensenuk: 👍

PR #283, opened 2018-10-11
Reduce initial RU value for un-paritioned collections

leewadhams:
Reducing the initial RU values for the collections that don't have a shard key.

We have a story for DevOps to make this configurable but talking to Steve yesterday he doesn't think it's going to be easy to do.

We'll therefore need to manually set the RU's in the environments. For our Dev ones we can just default to the lowest value to keep costs down.

### DIFF
--- a/src/Data/CosmosDb/Configuration.json
+++ b/src/Data/CosmosDb/Configuration.json
@@ -5,7 +5,7 @@
             "Collections": [
                 {
                     "CollectionName": "vacancies",
-                    "OfferThroughput": 1000,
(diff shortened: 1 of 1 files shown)


leewadhams:
@sachpatel yes we can. The issue is more around updating the values of existing collections. It negates the benefit a little of having multiple configs. Eg. having a prod one now wouldn't update the vacancies collection with a new value. We can discuss more with @smneal. 

I don't think it negates this change to default to the lowest RU values though.

PR #285, opened 2018-10-12
ER-652 Regenerates closed vacancies in query store

dev4valtech:


src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/Projections/PublishedVacancyProjectionService.cs
```diff
@@ -0,0 +1,78 @@
...
+            var programmesData = programmesTask.Result.Data;
+            var closedVacancies = closedVacanciesTask.Result.ToList();
+
+            var regenerateLiveVacanciesViewsTask = RegenerateLiveVacancies(liveVacancies, programmesData);
+            var regenerateClosedVacanciesViewsTask = RegenerateClosedVacancies(closedVacancies, programmesData);
+
+            await Task.WhenAll(regenerateClosedVacanciesViewsTask, regenerateLiveVacanciesViewsTask);
+        }
+
+        private Task RegenerateLiveVacancies(List<Vacancy> liveVacancies, List<ApprenticeshipProgramme> programmes)
```
sachpatel: Don't think you need to pass the args in as List, could be IEnumerable.

dev4valtech: I am using list to avoid double iteration as I need to log the count.

PR #286, opened 2018-10-12
Mongo - Projecting required properties

chrisjensenuk:
We have some generic functions in our repositories which allow the retrieval of a subset of the underlying document.  I noticed that the passing of this generic type down to the Mongo driver was only doing a client side translation so the whole document was being returned and then converted on the client.

I've updated these queries so it builds the projection from the passed in generic type so mongo/cosmos only returns the required data.


src/Shared/Recruit.Vacancies.Client/Infrastructure/Mongo/MongoDbCollectionBase.cs
```diff
@@ -43,6 +42,19 @@ protected IMongoCollection<T> GetCollection<T>()
             return collection;
         }
 
+        protected ProjectionDefinition<T> GetProjection<T>()
+        {
+            ProjectionDefinition<T> projection = null;
```
sachpatel: Would this cope with documents with nested complex types in properties? Also will we always need the "Id" field.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Repositories/MongoDbApplicationReviewRepository.cs
```diff
@@ -65,21 +65,25 @@ public Task UpdateAsync(ApplicationReview applicationReview)
 
         public async Task<List<T>> GetForEmployerAsync<T>(string employerAccountId)
         {
-            var filter = Builders<BsonDocument>.Filter.Eq(EmployerAccountId, employerAccountId);
-            var collection = GetCollection<BsonDocument>();
+            var filter = Builders<T>.Filter.Eq(EmployerAccountId, employerAccountId);
```
sachpatel: Shouldn't this repository and the VacancyRepository return the full document domain object and mapping done in the service layer to the summary types. I thought Repository classes are for domain objects (read/write) and anything else would be Provider classes.

If we know it is a specific DTO would it not be more efficient to write the projection because if we are that concerned it would be a very specific use case.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Mongo/MongoDbCollectionBase.cs
chrisjensenuk: Nope. This just does simple top level properties which is all we need for now.  Anything more complex you'd probably want to write your own projection for anyway.  As long as the type you use contains an Id field then it will be populated

src/Shared/Recruit.Vacancies.Client/Infrastructure/Repositories/MongoDbApplicationReviewRepository.cs
chrisjensenuk: Maybe. I'm just trying to make the existing functionality a bit more throughput friendly.

leewadhams: I agree with both of you :-)
I think this probably requires further discussion. I think what muddies the water is that we're using the repositories for queries not just when we're handling a create/update/delete and are getting an entity to update. 

@mgwilliam do you have any views on this?

mgwilliam: I'm inclined to separate the query and repository operations as there's a clear difference in intent:

* repository read/write methods are to support changing the data
* queries methods are for non-trivial reads that need to be current (ie. can't be stale)

... also noting that the query store is used for slow moving / complex reads or consolidating numerous data sources into a single doc/view.

So, given the above, I'd envisage the "projection methods" to move to the query interface from the repository interface.

Regarding how the query is implemented... has the mongo aggregation pipeline been considered? That would likely be more efficient as the hard work can be done at the data tier rather than pulling back into C#

Happy to discuss when we're all in the office if preferred?

chrisjensenuk: I'll split the repo interfaces into xxxRepository and xxxQuery.

I'll have a look if aggregates would be more appropriate for some of these query functions.

chrisjensenuk: In these cases there is nothing to be gained from using aggregates.

chrisjensenuk:
I've refactored the repo function that gets all the applications for all the vacancies for an employer.  It now uses an aggregate so it now only returns a count for each status type for each vacancy for an employer.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Repositories/MongoDbApplicationReviewRepository.cs
```diff
@@ -64,17 +64,29 @@ public Task UpdateAsync(ApplicationReview applicationReview)
...
-
-            var result = await RetryPolicy.ExecuteAsync(context => collection.FindAsync<T>(filter, options),
-                new Context(nameof(GetForEmployerAsync)));
+            var collection = GetCollection<ApplicationReview>();
 
-            return await result.ToListAsync();
+            var aggregate = collection.Aggregate()
+
+                .Match(r => r.EmployerAccountId == employerAccountId
+                            && r.IsWithdrawn != true)
```
sachpatel: `== false`

chrisjensenuk: This translates into `"isWithdrawn" : { "$ne" : true }` so handles if isWithdrawn is null.  I'll change it to use a non-lambda filter to save further confusion.

PR #295, opened 2018-10-17
ER-768 Updating Manage vacancy view to show submitted vacancy in new read only tab.

sachpatel:


src/Employer/Employer.Web/Controllers/VacancyViewController.cs
```diff
@@ -43,7 +43,7 @@ public async Task<IActionResult> DisplayFullVacancy(VacancyRouteModel vrm)
                 return HandleRedirectOfEditableVacancy(vacancy);
             }
 
-            if (vacancy.ApplicationMethod == ApplicationMethod.ThroughExternalApplicationSite || (vacancy.Status != VacancyStatus.Live && vacancy.Status != VacancyStatus.Closed))
+            if (vacancy.Status == VacancyStatus.Draft && vacancy.Status == VacancyStatus.Referred)
```
dev4valtech: I am bit confused here. First of all there is an error in if condition where we are checking for status in draft AND referred, there can ever be one status. So I thought may be we need an OR here. But then I couldn't make sense between vacancy.CanEdit and Vacancy being in one of these status. CanEdit is the same except that it checks for deletion. 
If the intention is to allow access to deleted vacancy then we need to convert if condition into a sensibly named boolean property just like we have CanEdit.

sachpatel: Good spot and suggestion. Will change.

src/Employer/Employer.Web/Orchestrators/VacancyViewOrchestrator.cs
```diff
@@ -43,6 +43,11 @@ public async Task<DisplayVacancyViewModel> GetFullVacancyDisplayViewModelAsync(V
         {
             switch (vacancy.Status)
             {
+                case VacancyStatus.Approved:
```
dev4valtech: Per Mark's comment on V2 channel, this should show as Live.

chrisjensenuk: I'm not sure the status ever gets set to `Approved` at the moment so doubt it will drop into here anyway.  It will in future though when we have deferred live dates.

PR #297, opened 2018-10-17
ER-721 Hides incomplete fields message when editing rejected vacancy

dev4valtech:
I have introduced couple of new tag helpers asp-show and asp-hide to replace asp-condition. These read a lot better on the page and also we can avoid creating negating booleans. I haven't applied it everywhere as it is a big undertaking, a tech debt task is open to address this refactoring.

### DIFF
--- a/src/Shared/Recruit.Shared.Web/TagHelpers/ConditionTagHelper.cs
+++ b/src/Shared/Recruit.Shared.Web/TagHelpers/ConditionTagHelper.cs
@@ -15,4 +15,32 @@ namespace Esfa.Recruit.Shared.Web.TagHelpers
             }
         }
     }
+
+    [HtmlTargetElement(Attributes = "asp-hide")]
+    public class HideTagHelper : TagHelper
+    {
+        public bool AspHide { get; set; }
+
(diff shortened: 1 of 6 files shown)


src/Shared/Recruit.Shared.Web/TagHelpers/ConditionTagHelper.cs
```diff
@@ -15,4 +15,32 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
...
+        {
+            if (AspHide)
+            {
+                output.SuppressOutput();
+            }
+        }
+    }
+
+    [HtmlTargetElement(Attributes = "asp-show")]
+    public class ShowTagHelper : TagHelper
```
sachpatel: I would think all the `asp-condition` should be replaced with this now and you would drop that tag helper.

dev4valtech: I spoke to @leewadhams  about making this change, we both agreed its going to have a huge impact so left it for later, as mentioned in my comment a tech debt task is in place and @mgwilliam is aware of it.

PR #303, opened 2018-10-19
ER-663 blocked employers

chrisjensenuk:
- Added In-Memory cache and our own `Cache.cs` abstraction wrapper around it
- As this 'blocking' is not part of normal system behaviour and more of a authorisation concern I have implemented as an MVC attribute/filter and have kept it outside of the Recruit Client.
- Added seeding json to populate `configuration` collection
- To add an employer account id just add a string to the array
`{
  "_id": "BlockedEmployers",
  "employerAccountIds": ["AAAAAA", "BBBBB"]
}`
- The blocked employers list will be cached once a day.  If we need a block to happen quicker than that we'll need to bounce the application.  Keeping it simple for now.

### DIFF
--- a/src/Employer/Employer.Web/Caching/Cache.cs
+++ b/src/Employer/Employer.Web/Caching/Cache.cs
@@ -1,0 +1,26 @@
+﻿using System;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Caching.Memory;
+
+namespace Esfa.Recruit.Employer.Web.Caching
+{
+    public class Cache : ICache
+    {
+        private readonly IMemoryCache _memoryCache;
+
+        public Cache(IMemoryCache memoryCache)
+        {
+            _memoryCache = memoryCache;
+        }
+
+        public Task<T> CacheAsideAsync<T>(string key, DateTime absoluteExpiration, Func<Task<T>> objectToCache)
+        {
+            return _memoryCache.GetOrCreate(key, entry =>
+            {
+                entry.AbsoluteExpiration = absoluteExpiration;
+
+                return objectToCache();
+            });
+        }
+    }
+}
--- a/src/Employer/Employer.Web/Caching/Cache.cs
+++ b/src/Employer/Employer.Web/Caching/Cache.cs
@@ -15,7 +15,7 @@ namespace Esfa.Recruit.Employer.Web.Caching
 
         public Task<T> CacheAsideAsync<T>(string key, DateTime absoluteExpiration, Func<Task<T>> objectToCache)
         {
-            return _memoryCache.GetOrCreate(key, entry =>
+            return _memoryCache.GetOrCreateAsync(key, entry =>
             {
                 entry.AbsoluteExpiration = absoluteExpiration;
 
--- a/src/Employer/Employer.Web/Extensions/HttpContextExtensions.cs
+++ b/src/Employer/Employer.Web/Extensions/HttpContextExtensions.cs
@@ -6,7 +6,7 @@ namespace Esfa.Recruit.Employer.Web.Extensions
 {
     public static class HttpContextExtensions
     {
-        public static async Task SignOutEmployerWebAsync(this HttpContext httpContext, bool signOutOidc = true)
+        public static async Task SignOutEmployerWebAsync(this HttpContext httpContext)
         {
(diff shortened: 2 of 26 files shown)


src/Employer/Employer.Web/Filters/CheckEmployerBlockedFilter.cs
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        public async Task OnActionExecutionAsync(
+            ActionExecutingContext context,
+            ActionExecutionDelegate next)
+        {
+            var blockedEmployerAccountIds = await _cache.CacheAsideAsync(
+                BlockedEmployersCacheKey, 
+                CacheAbsoluteExpiryTime, 
+                () =>  _blockedEmployersProvider.GetBlockedEmployerAccountIdsAsync());
```
sachpatel: async () =>  await _blockedEmployersProvider.GetBlockedEmployerAccountIdsAsync()

src/Employer/Employer.Web/Filters/CheckEmployerBlockedFilter.cs
```diff
@@ -0,0 +1,46 @@
...
+using Esfa.Recruit.Employer.Web.Configuration.Routing;
+using Esfa.Recruit.Employer.Web.Exceptions;
+using Esfa.Recruit.Vacancies.Client.Application.Providers;
+using Microsoft.AspNetCore.Mvc.Filters;
+
+namespace Esfa.Recruit.Employer.Web.Filters
+{
+    public class CheckEmployerBlockedFilter : IAsyncActionFilter
+    {
+        private const string BlockedEmployersCacheKey = "blockedEmployers";
```
sachpatel: Maybe stick this in a CacheKey constant file going forward.

src/Employer/Employer.Web/Extensions/HttpContextExtensions.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Authentication;
+
+namespace Esfa.Recruit.Employer.Web.Extensions
+{
+    public static class HttpContextExtensions
+    {
+        public static async Task SignOutEmployerWebAsync(this HttpContext httpContext, bool signOutOidc = true)
```
sachpatel: Unused parameter

src/Employer/Employer.Web/Caching/Cache.cs
```diff
@@ -0,0 +1,26 @@
...
+    public class Cache : ICache
+    {
+        private readonly IMemoryCache _memoryCache;
+
+        public Cache(IMemoryCache memoryCache)
+        {
+            _memoryCache = memoryCache;
+        }
+
+        public Task<T> CacheAsideAsync<T>(string key, DateTime absoluteExpiration, Func<Task<T>> objectToCache)
```
sachpatel: `async` all the way?

```
public async Task<T> CacheAsideAsync<T>(string key, DateTime absoluteExpiration, Func<Task<T>> objectToCache)
        {
            return await _memoryCache.GetOrCreateAsync(key, entry =>
            {
                entry.AbsoluteExpiration = absoluteExpiration;

                return objectToCache();
            });
        }
```

src/Employer/Employer.Web/Filters/CheckEmployerBlockedFilter.cs
chrisjensenuk: 👍

src/Employer/Employer.Web/Extensions/HttpContextExtensions.cs
chrisjensenuk: 👍 Good spot

src/Employer/Employer.Web/Caching/Cache.cs
chrisjensenuk: 👍 I'll use GetOrCreateAsync. But as I'm not doing anything with the result I can just pass through the task so I don't need to add await/async.

chrisjensenuk:
As we're caching now it probably isn't that costly to register this as a global filter and do it on every request.  I'll leave it for now as it satisfies the story and time is tight. I'll add a note to the tech debt to review.

PR #306, opened 2018-10-23
Simplify cloning of a vacancy

sachpatel:


leewadhams:
I see why you've done it, and I originally started doing this, but it doesn't feel right/safe to do this. We change what the vacancy reference points to completely which could be cause issue if you want to do anything with the original vacancy afterwards e.g. log messages. 

If you guys are happy with it then I will certainly get over it ;-)

chrisjensenuk:
@leewadhams I agree with your concerns. My preference is for creating a new object.

sachpatel:
This should all point to 👉  @mgwilliam . Let's discuss when we are all in office. The most suitable approach might be to map it into a new vacancy.

sachpatel:
Closing because of the discussion had.

PR #311, opened 2018-10-25
ER-766 - Legal entity agreement

chrisjensenuk:
Upgraded package `SFA.DAS.Account.Api.Client` to version 1.3.1331.

Added `LegalEntityAgreementService`. This is injected into orchestrators that need to know if a legal entity has a signed agreement.

- This service first checks the `EditVacancyInfo` projection to see if the `hasLegalEntityAgreement` is set to true. If `true` the journey continues as normal. (This will be the most common scenario)
- If `false` the service checks the Employer Service directly as the agreement could have been signed after the `EditVacancyInfo` projection was last rebuilt
- if the employer has signed an agreement then a command is fired for the jobs to rebuild the `EditVacancyInfo` projection (This will set  `hasLegalEntityAgreement` to true in the projection) and the journey continues as normal.
- if the employer does not have a signed agreement then the journey is diverted to one of the stop pages.

Also fixed an issue that sometimes happens on TEST in the ErrorController where TempData.Add was generating an "An item with the same key has already been added" error.

Removed unused folder `QueryStringModels`.

### DIFF
--- a/src/Employer/Employer.Web/Controllers/ErrorController.cs
+++ b/src/Employer/Employer.Web/Controllers/ErrorController.cs
@@ -60,7 +60,7 @@ namespace Esfa.Recruit.Employer.Web.Controllers
                 if (exception is InvalidStateException)
                 {
                     _logger.LogError(exception, "Exception on path: {route}", routeWhereExceptionOccurred);
-                    TempData[TempDataKeys.DashboardErrorMessage] = exception.Message;
+                    AddDashboardMessage(exception.Message);
                     return RedirectToRoute(RouteNames.Dashboard_Index_Get, new { EmployerAccountId = accountId });
                 }
 
@@ -107,6 +107,14 @@ namespace Esfa.Recruit.Employer.Web.Controllers
 
             Response.StatusCode = (int)HttpStatusCode.Unauthorized;
             return View(ViewNames.BlockedEmployer);
+        }
+
+        private void AddDashboardMessage(string message)
+        {
+            if(TempData.ContainsKey(TempDataKeys.DashboardErrorMessage))
+                _logger.LogError($"Dashboard message already set in {nameof(ErrorController)}. Existing message:{TempData[TempDataKeys.DashboardErrorMessage]}. New message:{message}");
+
+            TempData[TempDataKeys.DashboardErrorMessage] = message;
         }
     }
 }
--- a/src/Employer/Employer.Web/Controllers/ErrorController.cs
+++ b/src/Employer/Employer.Web/Controllers/ErrorController.cs
@@ -60,7 +60,7 @@ namespace Esfa.Recruit.Employer.Web.Controllers
                 if (exception is InvalidStateException)
                 {
                     _logger.LogError(exception, "Exception on path: {route}", routeWhereExceptionOccurred);
-                    TempData.Add(TempDataKeys.DashboardErrorMessage, exception.Message);
+                    TempData[TempDataKeys.DashboardErrorMessage] = exception.Message;
                     return RedirectToRoute(RouteNames.Dashboard_Index_Get, new { EmployerAccountId = accountId });
                 }
 
--- a/src/Employer/Employer.Web/Controllers/VacancyPreviewController.cs
+++ b/src/Employer/Employer.Web/Controllers/VacancyPreviewController.cs
@@ -55,6 +55,8 @@ namespace Esfa.Recruit.Employer.Web.Controllers
 
                 if (response.Data.HasLegalEntityAgreement == false)
                     return RedirectToRoute(RouteNames.LegalEntityAgreement_HardStop_Get);
+
+                throw new Exception("Unknown submit state");
             }
 
             var viewModel = await _orchestrator.GetVacancyPreviewViewModelAsync(m);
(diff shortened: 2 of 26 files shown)


src/Employer/Employer.Web/Services/LegalEntityAgreementService.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            var legalEntity = await GetLegalEntityAsync(employerAccountId, legalEntityId);
+
+            if(legalEntity == null)
+                return false;
+            
+            if (legalEntity.HasLegalEntityAgreement)
+                return true;
+
+            //Agreement may have been signed since the projection was created. Check Employer Service.
```
leewadhams: Don't know if this is possible but could there be a case where a contract is superseded by another and is yet to be signed?

src/Employer/Employer.Web/Employer.Web.csproj
```diff
@@ -9,31 +9,36 @@
...
-    <PackageReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Design" Version="2.1.4"/>
-    <PackageReference Include="NWebsec.AspNetCore.Middleware" Version="1.1.0"/>
-    <PackageReference Include="NLog.Web.AspNetCore" Version="4.5.4"/>
-    <PackageReference Include="NLog" Version="4.5.4"/>
-    <PackageReference Include="SFA.DAS.Account.Api.Client" Version="1.2.0.62882"/>
-    <PackageReference Include="SFA.DAS.NLog.Targets.Redis" Version="1.1.5"/>
-    <PackageReference Include="SFA.DAS.Providers.Api.Client" Version="0.10.138"/>
-    <PackageReference Include="SFA.DAS.Apprenticeships.Api.Types" Version="0.10.139"/>
-    <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.4.1"/>
+    <Compile Remove="QueryStringModels\**" />
```
leewadhams: Are these Visual Studio treats?

src/Employer/Employer.Web/Controllers/ErrorController.cs
```diff
@@ -60,7 +60,7 @@ public async Task<IActionResult> ErrorHandler()
                 if (exception is InvalidStateException)
                 {
                     _logger.LogError(exception, "Exception on path: {route}", routeWhereExceptionOccurred);
-                    TempData.Add(TempDataKeys.DashboardErrorMessage, exception.Message);
+                    TempData[TempDataKeys.DashboardErrorMessage] = exception.Message;
```
sachpatel: Is this hiding an issue? Could you log out what message is already there before overwriting so we can find the root issue. If it is multiple requests then something else has gone wrong.

src/Employer/Employer.Web/Employer.Web.csproj
chrisjensenuk: Its doing all sorts of weird things with this folder! I'll try and simplify

src/Employer/Employer.Web/Services/LegalEntityAgreementService.cs
chrisjensenuk: Yep this could happen. However at the moment we only need to check that at least one contract is signed regardless of the version. I've checked this with @mgwilliam

src/Employer/Employer.Web/Controllers/VacancyPreviewController.cs
```diff
@@ -50,7 +50,11 @@ public async Task<IActionResult> Submit(SubmitEditModel m)
 
             if (ModelState.IsValid)
             {
-                return RedirectToRoute(RouteNames.Submitted_Index_Get);
+                if (response.Data.IsSubmitted)
+                    return RedirectToRoute(RouteNames.Submitted_Index_Get);
+
+                if (response.Data.HasLegalEntityAgreement == false)
+                    return RedirectToRoute(RouteNames.LegalEntityAgreement_HardStop_Get);
```
dev4valtech: There is a flag redundancy as I see it, IsSubmitted and HasLegalAgreement they both reflect to the same cause... this can be changed to ... 

```suggestion
                if (response.Data.HasLegalEntityAgreement == false)
                    return RedirectToRoute(RouteNames.LegalEntityAgreement_HardStop_Get);
                
                return RedirectToRoute(RouteNames.Submitted_Index_Get);    
```

chrisjensenuk: My thinking is this is more explicit.

src/Employer/Employer.Web/Controllers/VacancyPreviewController.cs
```diff
@@ -55,6 +55,8 @@ public async Task<IActionResult> Submit(SubmitEditModel m)
 
                 if (response.Data.HasLegalEntityAgreement == false)
                     return RedirectToRoute(RouteNames.LegalEntityAgreement_HardStop_Get);
+
+                throw new Exception("Unknown submit state");
```
sachpatel: We have an `InvalidStateException` class

chrisjensenuk: An `InvalidStateException`  gets turned into a dashboard message saying a "colleague has made changes..." This exception should never fire.

PR #313, opened 2018-10-26
ER-797 - ARM template updates for non levy employer job

sachpatel:
If we can get this ARM template in, it would be okay as the next PR will have the code that uses it within the project/job.


sachpatel:
@jack-education This has been updated

PR #316, opened 2018-10-26
ER-790 Levy Declaration

leewadhams:
- Adds a new global filter to MVC Filter pipeline. Uses filter ordering to ensure its run after the Employer Blocked filter.
- Cookie is used to cache the fact that a user/account combination has selected being a levy payer already. Note: if the user has multiple account only the last one in stored in the cookie.


src/Employer/Employer.Web/Filters/LevyDeclarationCheckFilter.cs
```diff
@@ -0,0 +1,112 @@
...
+                    context.Result = new RedirectToRouteResult(RouteNames.LevyDeclaration_Get, new { employerAccountId });
+                    
+                    return;
+                }
+            }
+
+            await next(); 
+        }
+
+        private bool RequestIsForWhiteListedPage(ActionExecutingContext context)
```
chrisjensenuk: Would an Ignore attribute work here?
https://stackoverflow.com/questions/9953760/how-to-disable-a-global-filter-in-asp-net-mvc-selectively

src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/SaveUserLevyDeclarationCommandHandler.cs
```diff
@@ -0,0 +1,33 @@
...
+            _repository = repository;
+        }
+
+        public async Task Handle(SaveUserLevyDeclarationCommand message, CancellationToken cancellationToken)
+        {
+            _logger.LogInformation($"Updating user with id: {{userId}} with Legacy declaration of {message.DeclaringAsLevyEmployer}", message.UserId);
+            
+            var user = await _repository.GetAsync(message.UserId);
+
+            user.DeclaredAsLevyPayer = message.DeclaringAsLevyEmployer;
```
chrisjensenuk: I think a user can be attached to multiple employers so should this be a list of employer/bool combinations?

src/Employer/Employer.Web/Filters/LevyDeclarationCheckFilter.cs
leewadhams: We need to do more than skip in the circumstance where they have the cookie and are trying to navigate to the levy declaration page. We want to redirect to dashboard as opposed to skipping.

src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/SaveUserLevyDeclarationCommandHandler.cs
mgwilliam: it could just be a single cookie containing the employer account and user id couldn't it?
that way it's indicating who has been prompted... if the **current** user and/or employer doesn't match then that should trigger the prompt.

src/Employer/Employer.Web/Configuration/CookieNames.cs
```diff
@@ -8,5 +8,6 @@ public static class CookieNames
         public const string SetupEmployer = "setup-employer-{0}";
         public const string VacancyProposedClosingDate = "vacancy-closingDate-{0}";
         public const string VacancyProposedStartDate = "vacancy-startDate-{0}";
+        public const string LevyEmployerIndicator = "recruit-levy";
```
sachpatel: will need to put this on the MA cookies page.

src/Employer/Employer.Web/Filters/LevyDeclarationCheckFilter.cs
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        public int Order { get; } = 50;
+
+        public async Task OnActionExecutionAsync(ActionExecutingContext context, ActionExecutionDelegate next)
+        {
+            var employerAccountId = context.RouteData.Values[RouteValues.EmployerAccountId]?.ToString().ToUpper();
+            var userId = context.HttpContext.User.GetUserId();
+            
+            var haveValidCookie = HasValidLevyCookie(context, employerAccountId);
```
sachpatel: hasValidCookie?

src/Employer/Employer.Web/Orchestrators/LevyDeclarationOrchestrator.cs
```diff
@@ -0,0 +1,38 @@
...
+
+            return new LevySelectionOrchestratorResponse 
+            {
+                RedirectRouteName = viewModel.ConfirmAsLevyPayer.Value ? RouteNames.Dashboard_Index_Get : RouteNames.NonLevyInfo_Get,
+                CreateLevyCookie = viewModel.ConfirmAsLevyPayer.Value
+            };
+        }
+    }
+
+    public class LevySelectionOrchestratorResponse
```
sachpatel: Could we not use or extend off the Orchestrator response class?

src/Employer/Employer.Web/Services/LevyDeclarationCookieWriter.cs
```diff
@@ -0,0 +1,41 @@
...
+                return null;
+
+            return _dataProtector.Unprotect(content);
+        }
+
+        public void WriteCookie(HttpResponse response, string userId, string employerAccountId)
+        {
+            var protectedValue = _dataProtector.Protect($"{userId}-{employerAccountId}");
+
+            response.Cookies.Append(CookieNames.LevyEmployerIndicator, protectedValue, EsfaCookieOptions.GetDefaultHttpCookieOption(_hostingEnvironment));
```
sachpatel: Can you make sure we delete this cookie on signout if if it to be repopulated on sign in.

src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/SaveUserLevyDeclarationCommandHandler.cs
```diff
@@ -0,0 +1,39 @@
...
+            
+            var user = await _repository.GetAsync(message.UserId);
+
+            if (user.AccountsDeclaredAsLevyPayers.Contains(message.EmployerAccountId))
+            {
+                _logger.LogWarning($"The account {message.EmployerAccountId} was already in the list of declared levy payers for user: {message.UserId}");
+                return;
+            }
+
+            user.AccountsDeclaredAsLevyPayers .Add(message.EmployerAccountId);
```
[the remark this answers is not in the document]
leewadhams: 👍

src/Employer/Employer.Web/Services/LevyDeclarationCookieWriter.cs
sachpatel: I think we need to manage our cookies better. Not 💯 on this one being an issue. But any cookie that we do not define an expiry for, we should make sure we clean up. Some of the old dashboard filter cookies are hanging around on user machines and we want to remove the listing from the MA page.

src/Employer/Employer.Web/Filters/LevyDeclarationCheckFilter.cs
leewadhams: 👍

src/Employer/Employer.Web/Orchestrators/LevyDeclarationOrchestrator.cs
leewadhams: I'm not sure it sits very well Orchestrator Response class as that class is more focussed towards that validation of entity operation. Whereas what I need here is the choice of where to navigate the user to.

src/Employer/Employer.Web/Services/LevyDeclarationCookieWriter.cs
leewadhams: You're right. But in this scenario there's a benefit of not deleting/expiring. The functionality for selecting an option means that the cookie is always valid. i.e. they aren't able to go from selecting 'yes' to 'no'.

src/Shared/Recruit.Vacancies.Client/Domain/Entities/User.cs
```diff
@@ -1,15 +1,17 @@
...
+    public class User
+    {
+        public Guid Id { get; set; }
+        public string IdamsUserId { get; set; } 
+        public UserType UserType { get; set; }
+        public string Name { get; set; }
+        public string Email { get; set; }
+        public DateTime CreatedDate { get; set; }
+        public DateTime LastSignedInDate { get; set; }
+        public IList<string> AccountsDeclaredAsLevyPayers { get; set; } = new List<string>();
```
chrisjensenuk: We're only remembering accounts that have declared themselves as levy payers.  So if you are a non-levy payer then for each session you are always going to be asked the question about being a levy payer or not.  Is this right?

leewadhams: Yep that's correct. Any request to a page will redirect you to the choice again.

PR #317, opened 2018-10-26
Going down the wrong path for nothing

sachpatel:
This is pure guesswork and hoping the `OnRedirectToIdentityProviderForSignOut` is something that might help.

This actually worked for me after signing into test-eas and being on my localhost employer web on two different tabs and signing out from my recruit website and refreshing the MA dashboard on the other tab. The recruit tab took me to a page:

image

and refreshing the MA dashboard in the other tab took me to:

image


sachpatel:
After a long day, one thing is clear, and that is I am not thinking **clearly** 🌁

PR #318, opened 2018-10-29
ER-717 - Removing '/commitments/' from route

chrisjensenuk:
Sultan tested existing behaviour on TEST, PRE & PROD. The only environment that is working is PRE. Both TEST and PROD need to have `/commitments/` removed from the url in order to work. With this in mind I’m going to fix so it works on TEST and PROD and add an item to Tech Debt to get this working on PRE. (Ideally commitments PRE env needs to be updated rather than us adding extra config).


leewadhams:
I'm not sure this is correct. I think both PRE and PROD require the /commitments/ part of the url. The other environments don't.

chrisjensenuk:
@leewadhams Sultan's test on Friday says this is the case.  Could you get him to do it again and observe please?  Thanks

PR #321, opened 2018-10-29
ER-718 Implement SSO

leewadhams:
**Waiting for DevOps to restart the employerusers cloud services to allow testing**

On signout from recruit it will now redirect to the idams "You've logged out" page. This page contains iframes with the src for the signout of all the other subsites.

- Removed Redirect in logout as was blocking the required behaviour of oidc middleware
- Removed Logout from Blocked User story as this isn't required.

### DIFF
--- a/src/Employer/Employer.Web/Controllers/ErrorController.cs
+++ b/src/Employer/Employer.Web/Controllers/ErrorController.cs
@@ -82,7 +82,7 @@ namespace Esfa.Recruit.Employer.Web.Controllers
                 if (exception is BlockedEmployerException)
                 {
                     _logger.LogInformation($"{exception.Message}. Path: {routeWhereExceptionOccurred}");
-                    return await BlockedEmployerAsync();
+                    return BlockedEmployerAsync();
                 }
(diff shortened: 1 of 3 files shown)


src/Employer/Employer.Web/Controllers/ErrorController.cs
```diff
@@ -101,11 +101,10 @@ private IActionResult PageNotFound()
             return View(ViewNames.PageNotFound);
         }
 
-        private async Task<IActionResult> BlockedEmployerAsync()
+        private IActionResult BlockedEmployerAsync()
         {
-            await HttpContext.SignOutEmployerWebAsync();
```
chrisjensenuk: The blocked employer story said to explicitly to log out when an employer is blocked.

leewadhams: I checked with @mgwilliam  and Faizal and this behaviour is ok. Logging out of just recruit does make much sense and now using the SSO forces a redirect to an Idams logout page.

PR #324, opened 2018-10-29
Toggling web jobs

sachpatel:


src/Jobs/Recruit.Vacancies.Jobs/VacancyStatus/VacancyStatusJob.cs
```diff
@@ -15,16 +16,24 @@ public class VacancyStatusJob
         private const bool CanRunOnStartup = false;
 #endif
```
leewadhams: Do we need the compiler directives now?

https://github.com/SkillsFundingAgency/das-recruit/blob/00d9fbb45c6697e4a3a33a36d33e671e0c336001/src/Jobs/Recruit.Vacancies.Jobs/VacancyStatus/VacancyStatusJob.cs#L13-L17

src/Data/CosmosDb/Console.RecruitSeedDataWriter/SeedScript.ps1
```diff
@@ -13,6 +13,7 @@ try {
 	
 	& dotnet "$PSScriptRoot/Esfa.Recruit.Console.RecruitSeedDataWriter.dll" $($dbConnectionString) -c configuration -f "$PSScriptRoot/Data/BlockedEmployers.json"
```
leewadhams: Should this still be there??

sachpatel:
Happy to delay this is it is not priority and configure the on/off based on inheriting the trigger attributes. Will remove the preprocessor directives but maintain the running of the jobs on startup for local development.

PR #335, opened 2018-11-01
ER-814 minimum wages 2019

chrisjensenuk:
- Added 2019 wages
- Removed ValidTo from Minimum wage.  (Uses the start of the next period to determine end)
- Added the period year to the validation message


src/Data/CosmosDb/Console.RecruitSeedDataWriter/Data/MinimumWageRanges.json
```diff
@@ -4,10 +4,15 @@
   "ranges": [
     {
       "validFrom": ISODate("2018-04-01T00:00:00.000Z"),
```
leewadhams: Should we update the lastUpdateDate?

chrisjensenuk: 👍

PR #338, opened 2018-11-02
ER-812 - Explicitly setting status to 500

chrisjensenuk:
No rush to merge this.


chrisjensenuk:
As no exception is thrown then it doesn't trigger the handler again.

PR #339, opened 2018-11-05
ER-796 - Create vacancy from Dashboard and Submitted pages

chrisjensenuk:
Created a new 'create' endpoint that redirects to either the list of cloned vacancies or starts a new vacancy.

### DIFF
--- a/src/Employer/Employer.Web/Controllers/CreateVacancyOptionsController.cs
+++ b/src/Employer/Employer.Web/Controllers/CreateVacancyOptionsController.cs
@@ -23,6 +23,9 @@ namespace Esfa.Recruit.Employer.Web.Controllers
         {
             var vm = await _orchestrator.GetCreateOptionsViewModelAsync(employerAccountId);
 
+            if(vm.HasClonableVacancies == false)
+                return RedirectToRoute(@RouteNames.CreateVacancy_Get);
+
             return View(vm);
         }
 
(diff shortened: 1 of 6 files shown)


src/Employer/Employer.Web/Controllers/CreateVacancyOptionsController.cs
```diff
@@ -18,6 +18,16 @@ public CreateVacancyOptionsController(CreateVacancyOptionsOrchestrator orchestra
             _orchestrator = orchestrator;
         }
 
+        [HttpGet("create", Name = RouteNames.CreateVacancyOptions_Create_Get)]
+        public async Task<IActionResult> RedirectToCreateOrOptions([FromRoute]string employerAccountId)
+        {
+            var hasCloneableVacancies = await _orchestrator.HasCloneableVacancies(employerAccountId);
+
```
sachpatel: Should this not be performed on the `create-options` action as well?

chrisjensenuk: I don't think so. Whenever we want to create a vacancy we just need to point to the `create` endpoint  then this will work out where to redirect to.  the `create-options` endpoint will only ever by called by the redirect in `create`

sachpatel: what if `create-options` is typed in.

chrisjensenuk: You see the page with an empty list of cloneable vacancies.

sachpatel: For good practice I think it should redirect.

chrisjensenuk: Personally I don't have a problem with it but will add a redirect 👍

PR #343, opened 2018-11-05
Feature/ER-615 planned outage message functionality

sachpatel:
Has to modify the `asp-show` tag helper slightly.

image


sachpatel:
@chrisjensenuk 👍  Good suggestion. Will hold off for now because of minimum use of JS (apart from the training select drop down). Have added to a page of possible JS enhancements we would like to make - https://skillsfundingagency.atlassian.net/wiki/spaces/RAAV2/pages/872218925/Tech+enhancements

PR #344, opened 2018-11-05
ER-811 - Training provider input changed from number to text

chrisjensenuk:
This fixes a couple of issues:

In IE Edge the implementation of `type='number'` allows you to enter alphanumerics but prevents the form from being submitted.  We have code to disable submit buttons to prevent multiple form submits so the button gets disabled and never gets re-enabled.

The other issue is you can increment/decrement the number using up/down arrows & mouse scroll wheel. This is not useful behaviour with regard to UKPRNs.

So to fix this issue I'm changing the input from number to text. We have regex as part of server-side validation already to ensure numbers are used

### DIFF
--- a/src/Employer/Employer.Web/Controllers/Part2/TrainingProviderController.cs
+++ b/src/Employer/Employer.Web/Controllers/Part2/TrainingProviderController.cs
@@ -41,7 +41,10 @@ namespace Esfa.Recruit.Employer.Web.Controllers.Part2
                 return View(vm);
             }
 
-            var providerExists = await _orchestrator.ConfirmProviderExists(long.Parse(m.Ukprn));
+            if (long.TryParse(m.Ukprn, out var ukprnAsLong) == false)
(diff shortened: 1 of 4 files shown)


src/Employer/Employer.Web/Controllers/Part2/TrainingProviderController.cs
```diff
@@ -41,7 +41,10 @@ public async Task<IActionResult> SelectTrainingProvider(SelectTrainingProviderEd
                 return View(vm);
             }
 
-            var providerExists = await _orchestrator.ConfirmProviderExists(long.Parse(m.Ukprn));
+            if (long.TryParse(m.Ukprn, out var ukprnAsLong) == false)
+                return await ProviderNotFound(m);
```
chrisjensenuk: This IF statement is just belt-and-braces. The regex validation would prevent us getting this far.

PR #345, opened 2018-11-05
ER-820 Correct call to write levy cookie

leewadhams:
When writing the levy cookie after selecting the 'yes' option, the cookie contents was being written incorrectly. The result of this is that the next page they request will correctly say the cookie is invalid and remove it. It will then correctly regenerate the cookie. Subsequent requests will work as expected.


leewadhams:
Nope it's not 👍

PR #348, opened 2018-11-06
ER-651 - Moving away from open redirect 

sachpatel:
This is an enhancement to the item I did previously. Performing an open redirect based on the incoming `Referer` is bad practice as the referer could be an external site.

Happy to make changes if someone has a more appropriate solution.

### DIFF
--- a/src/Employer/Employer.Web/Startup.Configure.cs
+++ b/src/Employer/Employer.Web/Startup.Configure.cs
@@ -27,8 +27,7 @@ namespace Esfa.Recruit.Employer.Web
             
             if (env.IsDevelopment())
             {
-                //app.UseDeveloperExceptionPage();
-                app.UseExceptionHandler("/error/handle");
+                app.UseDeveloperExceptionPage();
             }
             else
             {
--- a/src/Employer/Employer.Web/Startup.Configure.cs
+++ b/src/Employer/Employer.Web/Startup.Configure.cs
@@ -27,7 +27,8 @@ namespace Esfa.Recruit.Employer.Web
             
             if (env.IsDevelopment())
             {
-                app.UseDeveloperExceptionPage();
+                //app.UseDeveloperExceptionPage();
+                app.UseExceptionHandler("/error/handle");
             }
             else
             {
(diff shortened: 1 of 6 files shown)


sachpatel:
@leewadhams Yes it should but I think this is a more clear way of performing the required redirect and it is a bit more explicit from within the controller code. I am happy to make any amendments or improvements if suggested?

leewadhams:
👍 Was just checking that we do have the safety net of the NWebSec check.

src/Employer/Employer.Web/Views/Shared/_Layout.cshtml
```diff
@@ -90,7 +90,7 @@ public string GetTitle()
 
     <div asp-show="@ViewContext.ViewData[ViewDataKeys.CanShowOutageMessage] as bool?" id="global-outage-message" class="global-message">
         <p>@ViewContext.ViewData[ViewDataKeys.PlannedOutageMessage]</p>
-        <p><a asp-route="@RouteNames.DismissOutageMessage_Get">Dismiss</a></p>
+        <p><a asp-route="@RouteNames.DismissOutageMessage_Get" asp-route-returnUrl="@ViewContext.HttpContext.Request.Path">Dismiss</a></p>
```
chrisjensenuk: Could we encrypt the returnUrl? Then if it decrypts OK then we know its one of ours and don't need to do the regex.

chrisjensenuk: If we turn this to a POST then we'd have the CSRF security of the Anti-forgery token.

sachpatel: Will change to use POST 👍

src/Employer/Employer.Web/Startup.Configure.cs
```diff
@@ -27,7 +27,8 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, IOptions
             
             if (env.IsDevelopment())
             {
-                app.UseDeveloperExceptionPage();
+                //app.UseDeveloperExceptionPage();
+                app.UseExceptionHandler("/error/handle");
```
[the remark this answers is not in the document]
sachpatel: 😬 will remove.

PR #350, opened 2018-11-06
ER-807 - Closing date extension

chrisjensenuk:
No longer clearing the cookies when Cancelling from the Edit Dates page.  Instead I'm ensuring cookies are cleared when we land on the Manage vacancy page.

Added handling of AggregateException in ErrorController. This is thrown from Task.WaitAll.

Refactored the Start/Closing date extending functionality to make the controller more skinny & move logic into orch as per other pages in employer Web.

Simplifying state handling by just throwing an InvalidStateException rather than marshalling redirects should a vacancy be in the wrong state.

### DIFF
--- a/src/Employer/Employer.Web/Controllers/ErrorController.cs
+++ b/src/Employer/Employer.Web/Controllers/ErrorController.cs
@@ -1,21 +1,17 @@
-﻿using System;
-using Microsoft.AspNetCore.Authorization;
-using Microsoft.AspNetCore.Mvc;
-using System.Diagnostics;
-using System.Linq;
-using Esfa.Recruit.Employer.Web.ViewModels.Error;
-using Microsoft.Extensions.Logging;
-using Microsoft.AspNetCore.Diagnostics;
-using Esfa.Recruit.Employer.Web.Configuration;
-using System.Net;
-using System.Threading.Tasks;
-using Esfa.Recruit.Vacancies.Client.Domain.Exceptions;
+﻿using Esfa.Recruit.Employer.Web.Configuration;
 using Esfa.Recruit.Employer.Web.Configuration.Routing;
 using Esfa.Recruit.Employer.Web.Exceptions;
-using Esfa.Recruit.Employer.Web.Extensions;
-using Esfa.Recruit.Employer.Web.Filters;
+using Esfa.Recruit.Employer.Web.ViewModels.Error;
+using Esfa.Recruit.Vacancies.Client.Domain.Exceptions;
 using Esfa.Recruit.Vacancies.Client.Infrastructure.Exceptions;
-using StackExchange.Redis;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Diagnostics;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Logging;
+using System;
+using System.Diagnostics;
+using System.Linq;
+using System.Net;
 
 namespace Esfa.Recruit.Employer.Web.Controllers
 {
--- a/src/Employer/Employer.Web/Controllers/ErrorController.cs
+++ b/src/Employer/Employer.Web/Controllers/ErrorController.cs
@@ -1,6 +1,8 @@
-﻿using Microsoft.AspNetCore.Authorization;
+﻿using System;
+using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using System.Diagnostics;
+using System.Linq;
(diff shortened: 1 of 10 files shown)


src/Employer/Employer.Web/Orchestrators/EditVacancyDatesOrchestrator.cs
```diff
@@ -40,15 +39,18 @@ public async Task<Vacancy> GetVacancyAsync(VacancyRouteModel vrm)
 
             Utility.CheckAuthorisedAccess(vacancy, vrm.EmployerAccountId);
 
+            if (vacancy.CanExtendStartAndClosingDates == false)
+                throw new InvalidStateException(string.Format(ViewModels.ErrorMessages.VacancyNotAvailableForEditing, vacancy.Title));
```
sachpatel: Would suggest a different error message such as `The dates of vacancy '{what's my name}' can't be edited.` as the current NotAvailableForEditing is for something slightly different of when the vacancy has been edited by another person. Maybe that one itself needs renaming.

src/Employer/Employer.Web/Orchestrators/EditVacancyDatesOrchestrator.cs
```diff
@@ -40,15 +39,18 @@ public async Task<Vacancy> GetVacancyAsync(VacancyRouteModel vrm)
...
         }
 
-        public Task<OrchestratorResponse<EditVacancyDatesViewModel>> GetEditVacancyDatesViewModel(VacancyRouteModel vrm, string proposedClosingDate, string proposedStartDate)
+        public async Task<OrchestratorResponse<EditVacancyDatesViewModel>> GetEditVacancyDatesViewModelAsync(VacancyRouteModel vrm, DateTime? proposedClosingDate, DateTime? proposedStartDate)
         {
             var vacancyTask = GetVacancyAsync(vrm);
             var programmesTask = _client.GetActiveApprenticeshipProgrammesAsync();
 
-            Task.WaitAll(vacancyTask, programmesTask);
+            await Task.WhenAll(vacancyTask, programmesTask);
```
sachpatel: 👍 - Is this where the aggregate exception was being thrown from?

src/Shared/Recruit.Vacancies.Client/Domain/Entities/Vacancy.cs
```diff
@@ -101,5 +101,10 @@ public class Vacancy
         public bool CanSendForReview => Status == VacancyStatus.Submitted && IsDeleted == false;
 
         public bool IsDisabilityConfident => DisabilityConfident == DisabilityConfident.Yes;
+
+        /// <summary>
+        /// We can extend the ClosingDate and StartDate for Live vacancies that have not been deleted
+        /// </summary>
+        public bool CanExtendStartAndClosingDates => Status == VacancyStatus.Live && IsDeleted == false;
```
sachpatel: Can a live vacancy be soft deleted? Would it not just be closed, otherwise we will have some lost orphan published vacancy known in the v2 system and published on FAA. This confuses the reasoning of the vacancy state/domain model.

src/Employer/Employer.Web/Controllers/ErrorController.cs
```diff
@@ -13,6 +15,7 @@
 using Esfa.Recruit.Employer.Web.Extensions;
 using Esfa.Recruit.Employer.Web.Filters;
 using Esfa.Recruit.Vacancies.Client.Infrastructure.Exceptions;
+using StackExchange.Redis;
```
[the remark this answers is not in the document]
chrisjensenuk: 👍 How did that get in there!

src/Employer/Employer.Web/Orchestrators/EditVacancyDatesOrchestrator.cs
chrisjensenuk: 👍

src/Employer/Employer.Web/Orchestrators/EditVacancyDatesOrchestrator.cs
chrisjensenuk: Yep. There are a few other WaitAlls in the code so I've changed this one. We may want to revisit the other ones at some point.

src/Shared/Recruit.Vacancies.Client/Domain/Entities/Vacancy.cs
chrisjensenuk: I agree we would never delete a live vacancy but we should always honour the IsDeleted flag.  Conceptually you should never be able to change any data on something that has been deleted.  This check is just belt-and-braces.

PR #351, opened 2018-11-06
ER-803 Fixed content display bug and removed feature toggle

dev4valtech:


src/Employer/Employer.Web/Orchestrators/Part2/ApplicationProcessOrchestrator.cs
```diff
@@ -67,20 +67,11 @@ public async Task<OrchestratorResponse> PostApplicationProcessEditModelAsync(App
         {
             var vacancy = await Utility.GetAuthorisedVacancyForEditAsync(_client, m, RouteNames.ApplicationProcess_Post);
 
-            if (_featureToggler.IsFeatureEnabled(FeatureNames.AllowThroughFaaApplicationMethod))
```
sachpatel: Shall we just remove the references and dependency of the `IFeature` feature toggle from this class and it's using statement.

src/Employer/Employer.Web/Views/Part2/ApplicationProcess/ApplicationProcess.cshtml
```diff
@@ -71,7 +56,7 @@
 </div>
 @section footerJS
 {
-    <feature name="@FeatureNames.AllowThroughFaaApplicationMethod">
+    <feature>
```
sachpatel: Probably remove this `feature` tag as it is redundant. Forget what I said earlier for reference we can just look back in history. It is intended to track feature/integration toggles on the following page https://skillsfundingagency.atlassian.net/wiki/spaces/RAAV2/pages/394068545/Feature+Toggles

src/Employer/Employer.Web/Orchestrators/Part2/ApplicationProcessOrchestrator.cs
chrisjensenuk: Yep. Get rid of using statement and ctor injection

PR #353, opened 2018-11-07
ER-720 Upgrade Webjob Sdk

leewadhams:
The format of *Program.cs* has changed quite a bit. It's now has a similar structure to asp.net core apps with the *Program.cs* file setting up the host.

The names of the Azure Storage connection strings have changed to fit in with the conventions that are used by the Sdk (Primarily for the webjobs dashboard and storage ones).


src/Jobs/Recruit.Vacancies.Jobs/Program.cs
```diff
@@ -34,152 +21,63 @@ static Program()
...
+                })
+                .UseConsoleLifetime();
+
+            var host = builder.Build();
+            using (host)
             {
-                jobConfiguration.DashboardConnectionString = null; // Reduces errors in output.
-                jobConfiguration.UseDevelopmentSettings();
-                jobConfiguration.Tracing.ConsoleLevel = TraceLevel.Off;
+                await host.RunAsync();
```
sachpatel: Just wondering if there was an uncaught exception triggering the job or some of the jobs that don't have try/catches would we not want the job host to catch, log and and maybe re-throw. Thinking of the cases of the `This job has met a horrible end` scenarios.

leewadhams: From what I can see any unhandled exceptions from the functions don't kill the host and logs out that that particular function has failed.

PR #357, opened 2018-11-12
ER-559 - Updating when the vacancy reference number assignation happens

sachpatel:
Triggering the vacancy reference number assignation on draft vacancy update and vacancy clone.

Added new `VacancyCloned` event to run the vacancy reference assignation on because it is possible for a user to clone a vacancy, not trigger an `UpdatedDraftVacancy` event by submitting the cloned vacancy straight away if it passes validation.

### DIFF
--- a/src/Jobs/Recruit.Vacancies.Jobs/DomainEvents/Handlers/Vacancy/DraftVacancyUpdatedHandler.cs
+++ b/src/Jobs/Recruit.Vacancies.Jobs/DomainEvents/Handlers/Vacancy/DraftVacancyUpdatedHandler.cs
@@ -25,7 +25,7 @@ namespace Esfa.Recruit.Vacancies.Jobs.DomainEvents.Handlers.Vacancy
             {
                 _logger.LogInformation($"Processing {nameof(DraftVacancyUpdatedEvent)} for vacancy: {{VacancyId}}", @event.VacancyId);
 
-                await _client.EnsureVacancyHasReferenceNumberAsync(@event.VacancyId);
+                await _client.AssignVacancyNumber(@event.VacancyId);
 
                 await _client.EnsureVacancyIsGeocodedAsync(@event.VacancyId);
 
--- a/src/Jobs/Recruit.Vacancies.Jobs/DomainEvents/Handlers/Vacancy/DraftVacancyUpdatedHandler.cs
+++ b/src/Jobs/Recruit.Vacancies.Jobs/DomainEvents/Handlers/Vacancy/DraftVacancyUpdatedHandler.cs
@@ -24,9 +24,11 @@ namespace Esfa.Recruit.Vacancies.Jobs.DomainEvents.Handlers.Vacancy
             try
             {
                 _logger.LogInformation($"Processing {nameof(DraftVacancyUpdatedEvent)} for vacancy: {{VacancyId}}", @event.VacancyId);
-               
+
+                await _client.EnsureVacancyHasReferenceNumberAsync(@event.VacancyId);
+
                 await _client.EnsureVacancyIsGeocodedAsync(@event.VacancyId);
-               
+
                 _logger.LogInformation($"Finished Processing {nameof(DraftVacancyUpdatedEvent)} for vacancy: {{VacancyId}}", @event.VacancyId);
             }
             catch (Exception ex)
@@ -37,4 +39,3 @@ namespace Esfa.Recruit.Vacancies.Jobs.DomainEvents.Handlers.Vacancy
         }
     }
 }
-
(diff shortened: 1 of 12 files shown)


src/Jobs/Recruit.Vacancies.Jobs/DomainEvents/Handlers/Vacancy/DraftVacancyUpdatedHandler.cs
```diff
@@ -24,9 +24,11 @@ public async Task HandleAsync(string eventPayload)
             try
             {
                 _logger.LogInformation($"Processing {nameof(DraftVacancyUpdatedEvent)} for vacancy: {{VacancyId}}", @event.VacancyId);
-               
+
+                await _client.EnsureVacancyHasReferenceNumberAsync(@event.VacancyId);
```
leewadhams: Are they awaited separately to avoid the vacancy overwrite issue?

src/Jobs/Recruit.Vacancies.Jobs/DomainEvents/Handlers/Vacancy/VacancyClonedDomainEventHandler.cs
```diff
@@ -6,12 +6,12 @@
 
 namespace Esfa.Recruit.Vacancies.Jobs.DomainEvents.Handlers.Vacancy
 {
-    public class VacancyCreatedHandler : DomainEventHandler, IDomainEventHandler<VacancyCreatedEvent>
+    public class VacancyClonedDomainEventHandler : DomainEventHandler, IDomainEventHandler<VacancyClonedEvent>
```
leewadhams: Do we not need a VacancyCreatedDomainEventHandler?

Is the @event also incorrect?

src/Shared/Recruit.Vacancies.Client/Infrastructure/Client/VacancyClient.cs
```diff
@@ -387,6 +387,16 @@ public Task WithdrawApplicationAsync(long vacancyReference, Guid candidateId)
             });
         }
 
+        public async Task EnsureVacancyHasReferenceNumberAsync(Guid vacancyId)
+        {
+            var vacancy = await _repository.GetVacancyAsync(vacancyId);
```
leewadhams: Do we need to retrieve the vacancy? Is this not just duplicate logic of what is in the command handler?

src/Jobs/Recruit.Vacancies.Jobs/DomainEvents/Handlers/Vacancy/DraftVacancyUpdatedHandler.cs
sachpatel: Yes that was my thinking.

leewadhams: 👍

src/Jobs/Recruit.Vacancies.Jobs/DomainEvents/Handlers/Vacancy/VacancyClonedDomainEventHandler.cs
sachpatel: So we still have a `VacancyCreated` event which is used by the `UpdateEmployerDashboardOnChange` errmmm in proc `EventHandler` 😕  but there is nothing out of process that we want to do on this event which is why we don't need it for a purpose. The event is still being stored in the domain events queue though just because of the nature of the event.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Client/VacancyClient.cs
sachpatel: This is for the purpose of the check to decide whether it needs to be assigned.

src/Jobs/Recruit.Vacancies.Jobs/DomainEvents/Handlers/Vacancy/VacancyClonedDomainEventHandler.cs
leewadhams: Why re-publish it at all? It blows up at the moment I think due to there being no handler registered for the event type.

This type is incorrect isn't it?

            var @event = DeserializeEvent<VacancyCreatedEvent>(eventPayload);

src/Shared/Recruit.Vacancies.Client/Infrastructure/Client/VacancyClient.cs
leewadhams: Isn't it just duplicating what's in the handler?

https://github.com/SkillsFundingAgency/das-recruit/blob/0e7546ee7d93ff7bd53aa0d811fd311744deb685/src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/AssignVacancyNumberCommandHandler.cs#L36-L41

src/Jobs/Recruit.Vacancies.Jobs/DomainEvents/Handlers/Vacancy/VacancyClonedDomainEventHandler.cs
sachpatel: 👍 Didn't notice that 🔫

PR #363, opened 2018-11-14
ER-826 Log performance counter for each command

dev4valtech:
Just logging performance counter as of now. We can use this for deciding on warnings threshold eventually.


src/Shared/Recruit.Vacancies.Client/Application/Aspects/RequestPerformanceBehaviour.cs
```diff
@@ -0,0 +1,37 @@
...
+        public RequestPerformanceBehaviour(ILogger<TRequest> logger)
+        {
+            _timer = new Stopwatch();
+
+            _logger = logger;
+        }
+
+        public async Task<TResponse> Handle(TRequest request, CancellationToken cancellationToken, RequestHandlerDelegate<TResponse> next)
+        {
+            _timer.Start();
```
sachpatel: can you not just use `var timer = Stopwatch.StartNew();` in this method instead of having a class level field?

src/Shared/Recruit.Vacancies.Client/Ioc/ServiceCollectionExtensions.cs
```diff
@@ -44,40 +44,31 @@
...
-                                                                return svc.GetAsync<QaRulesConfiguration>("QaRules").Result;
-                                                            });
-
-            services.RegisterStorageProviderDeps(configuration);
-
-            services.AddValidation();
-
-            services.AddRules();
+            services
+                .AddHttpClient()
```
sachpatel: ..... why?

src/Shared/Recruit.Vacancies.Client/Application/Aspects/RequestPerformanceBehaviour.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using MediatR;
+using Microsoft.Extensions.Logging;
+
+namespace Esfa.Recruit.Vacancies.Client.Application.Aspects
+{
+    public class RequestPerformanceBehaviour<TRequest, TResponse> : IPipelineBehavior<TRequest, TResponse>
+    {
+        private readonly Stopwatch _timer;
```
leewadhams: Does this need to be class level?

src/Shared/Recruit.Vacancies.Client/Application/Aspects/RequestPerformanceBehaviour.cs
```diff
@@ -0,0 +1,37 @@
...
+            _timer = new Stopwatch();
+
+            _logger = logger;
+        }
+
+        public async Task<TResponse> Handle(TRequest request, CancellationToken cancellationToken, RequestHandlerDelegate<TResponse> next)
+        {
+            _timer.Start();
+
+            var response = await next();
```
chrisjensenuk: What happens if the next handler throws an exception?  Do we need to wrap this in a try/catch ?

dev4valtech: I am presuming it will bubble up the error and we don't know how to treat the exception here.

chrisjensenuk: But we'd still want to write out the log though.  Catch, write the log, rethrow.

src/Shared/Recruit.Vacancies.Client/Ioc/ServiceCollectionExtensions.cs
dev4valtech: 🔨

src/Shared/Recruit.Vacancies.Client/Application/Aspects/RequestPerformanceBehaviour.cs
dev4valtech: The end game is this we want to determine if a particular request is slow performing, which means we only want to log the successful calls.

src/Shared/Recruit.Vacancies.Client/Application/Aspects/RequestPerformanceBehaviour.cs
dev4valtech: 🥇

PR #366, opened 2018-11-15
ER-785 - Logging CheckInfrastructure errors instead of exiting the app

chrisjensenuk:


src/QA/QA.Web/Program.cs
```diff
@@ -56,10 +56,17 @@ public static void Main(string[] args)
...
-            var collectionChecker = (MongoDbCollectionChecker)serviceProvider.GetService(typeof(MongoDbCollectionChecker));
-            collectionChecker.EnsureCollectionsExist();
+            try
+            {
+                var collectionChecker = (MongoDbCollectionChecker) serviceProvider.GetService(typeof(MongoDbCollectionChecker));
+                collectionChecker.EnsureCollectionsExist();
+            }
+            catch (Exception ex)
+            {
+                logger.Error(ex, "Error checking infrastructure");
```
sachpatel: Would we not want the message to be a little more specific about collections missing.

chrisjensenuk: That gets logged out in the ex message

PR #367, opened 2018-11-19
ER-831 - Redirect to blocked employer error page from action filter

sachpatel:
This is very crude and done to address the unhandled exception we are seeing on PROD.

Writing a custom middleware to intercept the exception was troublesome and I am not sure whether we need to throw an exception to intercept.

I also think it may be worth turning the `CheckEmployerBlockedFilter` into an `AuthorizationFilter`


leewadhams:
Out of interest what were the problems with the middleware option?

PR #370, opened 2018-11-23
Fix issue with startup of webjob

leewadhams:
Webjobs failed to start if the apprenticeship programmes were not populated in the collection.
This was due to the Singleton instance of the vacancy validator making a call to get them on webjob initialisation
Have made it a lazy call in that validator setup.


leewadhams:
@sachpatel this is primarily because we've moved the VacancyValidator to be a singleton and when we start-up the web job and configure ioc it creates the instance. Which then tries to get the apprenticeship programmes even before the ApprenticeshipProgrammesJob has had a chance to run and populate them.

PR #378, opened 2018-12-12
Add more CSP rules

leewadhams:


src/Employer/Employer.Web/Startup.Configure.cs
```diff
@@ -40,9 +40,18 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, IOptions
...
-                    .CustomSources("https://www.googletagmanager.com/",
-                                    "https://www.tagmanager.google.com/")
+                    {
+                        s.Self()
+                        .CustomSources("https://www.googletagmanager.com/",
+                                        "https://www.tagmanager.google.com/",
+                                        "https://tagmanager.google.com/",
+                                        "https://fonts.googleapis.com/");
+
+                        if (!env.IsProduction())
```
chrisjensenuk: Do we need this not-prod bit?

PR #395, opened 2019-01-08
Fix/ER-806 adding data protection storage provider to all three websites

sachpatel:
Also upgraded to using asp net core 2.2

This will require regression testing and monitoring of releases around the slot swapping to confirm anti-forgery token errors are not occurring.

### DIFF
--- a/src/Provider/Provider.Web/Startup.ConfigureServices.cs
+++ b/src/Provider/Provider.Web/Startup.ConfigureServices.cs
@@ -12,9 +12,9 @@ namespace Esfa.Recruit.Provider.Web
 {
     public partial class Startup
     {
-        private IConfiguration _configuration { get; }
-        private IHostingEnvironment _hostingEnvironment { get; }
-        private AuthenticationConfiguration _authConfig { get; }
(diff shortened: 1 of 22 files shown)


src/Provider/Provider.Web/Startup.ConfigureServices.cs
```diff
@@ -28,9 +30,13 @@ public void ConfigureServices(IServiceCollection services)
         {
             services.AddIoC(_configuration);
 
+            var redis = ConnectionMultiplexer.Connect(_configuration.GetConnectionString("Redis"));
+            services.AddDataProtection().PersistKeysToStackExchangeRedis(redis, "DataProtection-Keys");
+
             // Routing has to come before adding Mvc
             services.AddRouting(opt =>
             {
+                opt.LowercaseUrls = true;
```
dev4valtech: Perhaps make this consistent across other websites

sachpatel: don't want employer account codes to be lowercased on employer website.

PR #396, opened 2019-01-08
ER-858 html editor

chrisjensenuk:
Added TinyMCE editor to allow the addition of bullet points.

Added OWASP recommended `HTMLSanitizer` https://github.com/mganss/HtmlSanitizer to help protect against XSS scripting attacks.  This works by whitelisting allowed tags and attributes. These tags have been whitelisted `<p>,<br>,<ul>,<li>` all other tags and attributes will be stripped.

I've added a `HTMLValidator` to fluently validate HTML input.  This also uses `HTMLSanitizer` to check the input has been sanitised.

I've increased the server-side maxlength from 500 to 1000 chars to accommodate any extra mark-up. client side this is still set at 500 char limit (does not include mark-up in calculation).

HTML input is sanitised immediately before any validation is done. This means if any dodgy HTML has been posted it will just get stripped and not raise a validation error.  My thinking is we currently automatically strip whitespace from the beginning and end of input strings without raising a validation error so in a similar fashion I'm just stripping out non-whitelisted markup.

My only slight concern is that `HTMLSanitizer` is being called twice per HTML input posting.  The first time is done before setting the Domain entity property.  The second is done during validation as an insurance policy to ensure the value was sanitised in the first place.

I'm not sure about the performance of `HtmlSanitizer` yet until we get it into the environments but I'm logging out the duration of the sanitisation so we can track it.


src/Employer/Employer.Web/wwwroot/javascripts/application.js
```diff
@@ -174,6 +182,48 @@ function inViewport($el) {
...
+            var element = tinyMceEditor.getElement();
+            
+            tinyMceEditor.on('keyup',
+                function (e) {
+                    setEditorMaxLength(element, tinyMceEditor);
+                });
+            tinyMceEditor.on('focus',
+                function (e) {
+                    tinyMceEditor.editorContainer.classList.add("editor-focus");
+                    console.log("focus");
```
leewadhams: Was this logging just for debugging purposes?

src/Employer/Employer.Web/wwwroot/javascripts/application.js
```diff
@@ -174,6 +182,48 @@ function inViewport($el) {
...
+                });
+            tinyMceEditor.on('blur',
+                function (e) {
+                    tinyMceEditor.editorContainer.classList.remove("editor-focus");
+                    console.log("blur");
```
leewadhams: Was this logging just for debugging purposes?

src/Employer/Employer.Web/Views/Shared/_Layout.cshtml
```diff
@@ -197,6 +197,7 @@ public bool IsErrorPage()
     <script src="/lib/esfa-govuk/assets/javascripts/govuk-template.js"></script>
     <script src="/lib/esfa-govuk/assets/javascripts/govuk/show-hide-content.js"></script>
     <script src="/lib/esfa-stacker/stacker.js"></script>
+    <script src="/lib/tinymce-4.9.2/tinymce.min.js"></script>
```
leewadhams: Does it need to be included on every page?

leewadhams: Does this need to be on every page? Is it the dependency that application.js has on it which requires it here?

src/Employer/Employer.Web/wwwroot/javascripts/application.js
chrisjensenuk: Ah yes. Will remove.

src/Employer/Employer.Web/wwwroot/javascripts/application.js
chrisjensenuk: 👍

src/Employer/Employer.Web/Views/Shared/_Layout.cshtml
chrisjensenuk: 👍 I'll only reference it in the required pages. application.js is happy enough as long as we don't call initializeEditor without the script being referenced.

chrisjensenuk:
I've added TinyMCE to Provider.  Its not needed in QA so have left this for now.

PR #400, opened 2019-01-09
ER-835 Adds School to banned phrases

dev4valtech:
Not sure why some of the words in the list starts with uppercase letter and some don't, do we need to make it consistent by having all the words in lowercase?


dev4valtech:
@sachpatel no particular reason, but I would rather have it all in lower case than mixed to avoid any confusion

PR #412, opened 2019-01-14
ER-858 - Html validation improvements

chrisjensenuk:
- Only running the sanitizer as part of validation and not on post.
- If the sanitization process only returns HTML whitespace then fail validation


src/Shared/Recruit.Vacancies.Client/Application/Validation/Fluent/CustomValidators/HtmlValidator.cs
```diff
@@ -14,24 +14,20 @@ internal class HtmlValidator : PropertyValidator
...
-            if (string.IsNullOrWhiteSpace(value))
-                return true;
-
             var isHtmlValid = _sanitizer.IsValid(value);
 
             if (isHtmlValid == false)
                 return false;
 
-            if (!_regex.IsMatch(value))
+            if (string.IsNullOrWhiteSpace(value) == false &&
```
dev4valtech: This is fine, just a concern on optional fields, would this fail validation on optional fields where html is allowed and the value is null?

chrisjensenuk: As it stands yes. This can be adapted later to handle optional fields when we have them.  We could pass in a bool when we attach the validator to manage this.

PR #421, opened 2019-01-17
ER-669 Allows QA to access closed reviews

dev4valtech:
- Moved some of the review summary related stuff in to shared so I can reuse it on QA
- Created a copy of the review summary view as it is slightly different compared to rejected vacancy view on employer. 
- A lot of these changes  are due to namespace update and method name correction.


src/QA/QA.Web/Views/Dashboard/Index.cshtml
```diff
@@ -74,7 +74,7 @@
                 <td>@vacancy.EmployerName</td>
                 <td>@vacancy.VacancyTitle</td>
                 <td><span>VAC</span>@vacancy.VacancyReference</td>
-                <td>@vacancy.SubmittedDate.AsGdsDateTime()</td>
+                <td><span asp-hide="vacancy.IsClosed">With employer</span><span asp-hide="vacancy.IsClosed">@vacancy.SubmittedDate.AsGdsDateTime()</span></td>
```
chrisjensenuk: Should both of these be asp-hide?

src/Shared/Recruit.Vacancies.Client/Domain/Repositories/IVacancyReviewQuery.cs
```diff
@@ -9,7 +9,7 @@ public interface IVacancyReviewQuery
     {
         Task<List<T>> GetActiveAsync<T>();
         Task<List<VacancyReview>> GetForVacancyAsync(long vacancyReference);
-        Task<List<VacancyReview>> SearchAsync(long vacancyReference);
+        Task<VacancyReview> SearchByReferenceAsync(long vacancyReference);
```
chrisjensenuk: There can be many reviews for a single VacancyReference so should this be renamed to describe the review it is retrieving? `GetLatestReviewByReference`? Or something..

src/QA/QA.Web/Views/Dashboard/Index.cshtml
dev4valtech: Good catch... however this was not intended here... reverting back to original.

src/QA/QA.Web/Mappings/ReviewMapper.cs
```diff
@@ -211,6 +213,15 @@ public async Task<ReviewViewModel> Map(VacancyReview review)
             return vm;
         }
 
+        private string GetPageTitle(ReviewHistoriesViewModel historiesVm, Guid reviewId, ManualQaOutcome? reviewManualOutcome)
+        {
+            var timeFrame = historiesVm.Items.First().ReviewId == reviewId ? "Latest" : "Historical";
+            var outcome = reviewManualOutcome.GetValueOrDefault().ToString().ToLower();
```
chrisjensenuk: This is fine but maybe we could add this to EnumExtensions.GetDisplayName()

dev4valtech: Enum.ToString() does the job, the extension will not add any further value.

PR #422, opened 2019-01-18
ER-861 Creates vacancyReviews collection

dev4valtech:


sachpatel:
Also don't bump the patch and use +semver: skip

We will work out what version segment to bump when we do the migration.

PR #425, opened 2019-01-21
ER-669 Closed reviews shouldn't have assignment note

dev4valtech:
These flags on review model are so confusing... I have applied a patch here... going to add a tech debt to fix these flags properly.

### DIFF
--- a/src/QA/QA.Web/ViewModels/VacancyReviewSearchResultViewModel.cs
+++ b/src/QA/QA.Web/ViewModels/VacancyReviewSearchResultViewModel.cs
@@ -23,7 +23,7 @@ namespace Esfa.Recruit.Qa.Web.ViewModels
 
         public bool IsClosed { get; set; }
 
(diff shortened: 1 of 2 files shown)


src/QA/QA.Web/ViewModels/VacancyReviewSearchResultViewModel.cs
```diff
@@ -23,7 +23,7 @@ public class VacancyReviewSearchResultViewModel
 
         public bool IsClosed { get; set; }
 
-        public bool IsNotAvailableForReview => !IsAvailableForReview;
+        public bool IsNotAvailableForReview => !IsAvailableForReview && !IsClosed;
```
chrisjensenuk: IsNotAvailableForReview should be just be a simple inverse of IsAvailableForReview.  We shouldn't add extra bits to it as it'll no longer be the case.  Whatever is setting IsAvailableForReview needs to take into account if the review is closed.

PR #429, opened 2019-01-24
ER-880 Part 2 Skills

chrisjensenuk:
- I've  created a `SkillsViewModelHelper` class to share logic between Employer and Provider Skills orchestrators.
- I've removed redundant code that used to remove custom skills.  Now we display custom skills as checkboxes then this code was never called.
- Created a UnitTest project for Provider.Web


src/Provider/Provider.Web/ViewModels/Part2/Skills/SkillsViewModel.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Collections.Generic;
+using Esfa.Recruit.Shared.Web.ViewModels;
+using Esfa.Recruit.Shared.Web.ViewModels.Skills;
+
+namespace Esfa.Recruit.Provider.Web.ViewModels.Part2.Skills
+{
+    public class SkillsViewModel : SkillsEditModel, ISkillsViewModel
```
sachpatel: We have been trying to avoid this. ViewModels and EditModels are two separate concerns that don't go together and ViewModels should not inherit from EditModels. Otherwise we might as well use one class for input and output and use [BindNever] attributes for ignoring parts of request binding.

src/Shared/Recruit.Shared.Web/ViewModels/Skills/SkillsViewModelHelper.cs
```diff
@@ -0,0 +1,152 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
```
sachpatel: If the part and functionality within the orchestrator is really common, would it not be better if we shared those bits of the orchestrator that the orchestrator in the web projects delegate to? Might mean we don't have to have two sets of similar unit tests.

Same thing for the ViewModel and EditModel where you have created interfaces, would they not be better off having base classes. This might make the concrete classes in the web projects hollow but would still allow extensibility.

src/Shared/Recruit.Shared.Web/ViewModels/Skills/ISkillsEditModel.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Collections.Generic;
+
+namespace Esfa.Recruit.Shared.Web.ViewModels.Skills
+{
+    public interface ISkillsEditModel
```
sachpatel: We have a couple of interfaces like this that we would like to remove. They are not contracts for behaviour but state and are best avoided around dto's. Can we not use an abstract class instead.

src/Shared/Recruit.Shared.Web/ViewModels/Skills/SkillsViewModelHelper.cs
```diff
@@ -0,0 +1,152 @@
...
+        private const int ColumnOneCutOffIndex = 9;
+        private const char SortPrefixSeparator = '-';
+
+        private readonly Lazy<List<string>> _lazyCandidateSkills;
+        
+        private List<string> CandidateSkills => _lazyCandidateSkills.Value;
+        private IEnumerable<string> Column1BuiltInSkills => CandidateSkills.Take(ColumnOneCutOffIndex);
+        private IEnumerable<string> Column2BuiltInSkills => CandidateSkills.Skip(ColumnOneCutOffIndex);
+
+        public SkillsViewModelHelper(IRecruitVacancyClient vacancyClient)
```
sachpatel: Could we not use a passed in func delegate to avoid having this class knowing about the VacancyClient.

src/Provider/Provider.Web/ViewModels/Part2/Skills/SkillsViewModel.cs
chrisjensenuk: This is just cut and pasted from exiting code. Most of our View Models inherit from Edit Models.

src/Shared/Recruit.Shared.Web/ViewModels/Skills/SkillsViewModelHelper.cs
chrisjensenuk: I'll have a look and see if we can get rid of the interface.  The problem is the VMs inherit from VacancyRouteModel which is different between Provider and Employer.  I'll have another think.

I didn't want to share too much of the orchestrator code to get away from tight coupling between Provider and Employer. The building of the skills array is the bulk of the code in the orchestrator and I thought that was more shareable - E.g. When we build the API we can reuse this helper there too.

src/Shared/Recruit.Shared.Web/ViewModels/Skills/ISkillsEditModel.cs
chrisjensenuk: The differing VacancyRouteModels prevent this.  I'll have another rethink though and see if I can get rid of these interfaces.

src/Shared/Recruit.Shared.Web/ViewModels/Skills/SkillsViewModelHelper.cs
chrisjensenuk: 👍

src/Provider/Provider.Web/ViewModels/Part2/Skills/SkillsViewModel.cs
chrisjensenuk: I've removed the inheritance between the View and Edit model

src/Shared/Recruit.Shared.Web/ViewModels/Skills/ISkillsEditModel.cs
chrisjensenuk: I've removed the inheritance from VRM

PR #433, opened 2019-01-28
Defensive code around framework and standards job

sachpatel:
I have consolidated some of the name-spacing for unit test around vacancy client to avoid having `Vacancy.Client` appearing two times in the namespace declaration.


sachpatel:
@chrisjensenuk I think defensive against dupes is more appropriate for us than erroring. In an ideal world we wouldn't have to even dedupe but I feel funny about trusting what happens on that API as we haven't even found the root cause of why it went wrong.

👍 Will add a guard and logging around potential empty upsert call.

PR #438, opened 2019-01-30
ER-842 - Removing MinimumWageRanges.json from SeedScript

chrisjensenuk:
This is a clean-up ticket.  Even though I've deleted the MinimumWageRanges.json file I neglected to remove the reference from here.

The deployment script handles this missing file nicely so I don't think its worth redeploying the release for.


chrisjensenuk:
Closing PR as change already made in #442

PR #443, opened 2019-01-31
ER-875 Implements provider step 1 location

dev4valtech:
This one is identical to Employer's but not exact match. 
- The route used is `location` as `employer` is used for the first step. 
- No checks for soft/hard stop on legal entities selection as this is assumed. 
- Employer name will **not** be updated with legal entity name
- Some styling is missing, waiting on Richard to send me prototype


src/Shared/Recruit.Vacancies.Client/Infrastructure/QueryStore/QueryStoreClient.cs
```diff
@@ -96,17 +97,66 @@ public Task<EditVacancyInfo> GetEmployerVacancyDataAsync(string employerAccountI
 
         public Task<ProviderEditVacancyInfo> GetProviderVacancyDataAsync(long ukprn)
         {
+            //TODO get data from QueryStore Collection
```
sachpatel: 💥 We shouldn't need TODO's, the code/signatures and execution will tell us what is missing.

PR #446, opened 2019-02-01
ER-876 Implements important dates and training step in part1

dev4valtech:
This is exact copy of employers

### DIFF
--- a/src/Shared/Recruit.Vacancies.Client/Infrastructure/Client/IProviderVacancyClient.cs
+++ b/src/Shared/Recruit.Vacancies.Client/Infrastructure/Client/IProviderVacancyClient.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Threading.Tasks;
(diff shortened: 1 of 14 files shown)


src/Shared/Recruit.Vacancies.Client/Infrastructure/Client/IProviderVacancyClient.cs
```diff
@@ -15,5 +16,6 @@ public interface IProviderVacancyClient
         Task UpdateDraftVacancyAsync(Vacancy vacancy, VacancyUser user);
         Task<ProviderEditVacancyInfo> GetProviderEditVacancyInfoAsync(long ukprn);
         Task<EmployerInfo> GetProviderEmployerVacancyDataAsync(long ukprn, string employerAccountId);
+        Task<IEnumerable<IApprenticeshipProgramme>> GetActiveApprenticeshipProgrammesAsync();
```
chrisjensenuk: I think this should be in `IRecruitVacancyClient` and Employer and Provider should point to it.

PR #453, opened 2019-02-05
ER-868 - QA provider/employer stats

chrisjensenuk:
I have also removed `employerAccountId` from `vacancyReview` as it is never used and we should use the snapshot to get this information if we need it.


src/Data/Recruit.ChangeScripts/090-vacancyReviews_removing_employerAccountId.js
```diff
@@ -0,0 +1,53 @@
+{
```
dev4valtech: Since there is a migration planned in the next release (from `vacancyReviews` to `reviews`) we need to make sure in which order this script gets executed.

dev4valtech: This file is also included in the PR for ER-887

PR #454, opened 2019-02-05
ER-887 - QA Provider review

chrisjensenuk:
This follows on from ER-935 so contains commits from this PR.

### DIFF
--- a/src/QA/QA.Web/ViewModels/ReviewViewModel.cs
+++ b/src/QA/QA.Web/ViewModels/ReviewViewModel.cs
@@ -30,12 +30,12 @@ namespace Esfa.Recruit.Qa.Web.ViewModels
         public string NumberOfPositionsCaption { get; internal set; }
         public string OutcomeDescription { get; internal set; }
         public string HoursPerWeek { get; internal set; }
+        public OwnerType OwnerType { get; internal set; }
         public string PossibleStartDate { get; internal set; }
         public string ProviderContactName { get; internal set; }
         public string ProviderContactEmail { get; internal set; }
         public string ProviderContactTelephone { get; internal set; }
         public string ProviderName { get; internal set; }
-        public SourceOrigin SourceOrigin { get; internal set; }
         public string ThingsToConsider { get; internal set; }
         public string TrainingDescription { get; internal set; }
         public string VacancyDescription { get; internal set; }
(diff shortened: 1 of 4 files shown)


src/Data/Recruit.ChangeScripts/090-vacancyReviews_removing_employerAccountId.js
```diff
@@ -0,0 +1,53 @@
+{
```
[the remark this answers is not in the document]
sachpatel: how about we kill two birds with one stone and do the field drop when we do the migration? We will be taking the system down where we can run the two procedures one after the other.

src/Shared/Recruit.Vacancies.Client/Domain/Entities/VacancyReview.cs
```diff
@@ -30,7 +30,6 @@ public class VacancyReview
         public string ManualQaComment { get; set; }
         public List<ManualQaFieldIndicator> ManualQaFieldIndicators { get; set; }
         public string PrivateReviewNotes { get; set; }
```
sachpatel: Not against the dropping of this field if you have found no code relies on it but could you give a heads up in the future regarding this as it has implications on queries used for investigation purposes.

src/QA/QA.Web/ViewModels/ReviewViewModel.cs
```diff
@@ -31,7 +31,11 @@ public class
         public string OutcomeDescription { get; internal set; }
         public string HoursPerWeek { get; internal set; }
         public string PossibleStartDate { get; internal set; }
+        public string ProviderContactName { get; internal set; }
+        public string ProviderContactEmail { get; internal set; }
+        public string ProviderContactTelephone { get; internal set; }
         public string ProviderName { get; internal set; }
+        public SourceOrigin SourceOrigin { get; internal set; }
```
sachpatel: I don't think we need the SourceOrigin on this ViewModel to work out `IsEmployerVacancy`, could you please use the `OwnerType` property from the vacancy instead please?

src/Shared/Recruit.Vacancies.Client/Domain/Entities/VacancyReview.cs
sachpatel: What has Mark got to do with it

PR #461, opened 2019-02-08
ER-904_Migrate From Query Store to Table Storage.

shomavg:


src/Shared/Recruit.Vacancies.Client/Infrastructure/TableStore/StorageTableNames.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Esfa.Recruit.Vacancies.Client.Infrastructure.TableStore
+{
+    internal static class StorageTableNames
+    {
+        internal const string TableName = "QueryStore";
```
dev4valtech: Name should indicate underline table name ... perhaps ... `StorageTableNames.QueryStoreName`

src/Shared/Recruit.Vacancies.Client/Infrastructure/Client/VacancyClient.cs
```diff
@@ -82,8 +82,7 @@ public partial class VacancyClient : IRecruitVacancyClient, IEmployerVacancyClie
 
         public Task UpdateDraftVacancyAsync(Vacancy vacancy, VacancyUser user)
         {
-            var command = new UpdateDraftVacancyCommand
-            {
+            var command = new UpdateDraftVacancyCommand {
```
dev4valtech: Inconsistent formatting

src/Shared/Recruit.Vacancies.Client/Infrastructure/TableStore/StorageRetryPolicy.cs
```diff
@@ -0,0 +1,26 @@
...
+using Polly;
+using Polly.Retry;
+
+namespace Esfa.Recruit.Vacancies.Client.Infrastructure.TableStore
+{
+    public static class StorageRetryPolicy
+    {
+        public static RetryPolicy GetRetryPolicy(ILogger logger)
+        {
+            return Policy
```
dev4valtech: Wider question... I believe we are not using the azure inbuilt retry policies for consistency purpose?

shomavg: Yes. Since Polly is being used elsewhere in the solution, i too have used polly. I got this confirmed from Mark.

src/Shared/Recruit.Vacancies.Client/Infrastructure/TableStore/TableStorageQueryStore.cs
```diff
@@ -0,0 +1,125 @@
...
+            : base(logger, details)
+        {
+            var storageAccount = CloudStorageAccount.Parse(details.Value.ConnectionString);
+            TableClient = storageAccount.CreateCloudTableClient();
+            CloudTable = TableClient.GetTableReference(StorageTableNames.TableName);
+        }
+
+        public async Task<T> GetAsync<T>(string typeName, string key) where T : QueryProjectionBase
+        {
+            var retrieveOperation = TableOperation.Retrieve<QueryEntity>(typeName, key);
```
dev4valtech: Instead of TableOperation, why not use TableQuery which has generic queries that can translate response straight into custom object type ?

shomavg: TableOperation has a standard set of methods to perform various operations on a table like insert,replace etc.

src/Shared/Recruit.Vacancies.Client/Infrastructure/TableStore/TableStorageConnectionsDetails.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Esfa.Recruit.Vacancies.Client.Infrastructure.TableStore
+{
+    public class TableStorageConnectionsDetails
```
chrisjensenuk: TableStorageConfiguration maybe?

src/Shared/Recruit.Vacancies.Client/Infrastructure/TableStore/TableStorageQueryStore.cs
```diff
@@ -0,0 +1,125 @@
...
+            {
+                var actualItem = JsonConvert.DeserializeObject<T>(queryEntity.JsonData);
+                return actualItem;
+            }
+            return null;
+        }
+
+        public Task UpsertAsync<T>(T item) where T : QueryProjectionBase
+        {
+            CheckIfTableExists();
```
chrisjensenuk: Do we need to do this check on every operation? Can we create this table as part of the deployment process like we do with the Cosmos Mongo collections and just presume the table does exist?

PR #463, opened 2019-02-12
ER-990 - Application reviews aggregate query

sachpatel:


src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/VacancySummariesProvider/VacancySummariesProvider.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            var mongoQuery = pipeline.ToJson();
+
+            var db = GetDatabase();
+            var collection = db.GetCollection<BsonDocument>(MongoDbCollectionNames.Vacancies);
+
+            var aggResults = await RetryPolicy.ExecuteAsync(async context => await collection.AggregateAsync<VacancySummaryAggQueryResponseDto>(pipeline),
+                                                            new Context(nameof(RunAggPipelineQuery)));
+
+            var vacancySummaries = await aggResults.ToListAsync();
```
chrisjensenuk: The `ToListAsync()` needs to be inside the retry as this is the bit that does the fetching from the DB so can we get cursor errors here.  I'm updating other instances of this as part of ER-898.

src/Jobs/Recruit.Vacancies.Jobs/SpikeJob/SpikeJob.cs
```diff
@@ -0,0 +1,34 @@
+using System;
```
chrisjensenuk: Do you want to include this file?

src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/VacancySummariesProvider/VacancySummariesProvider.cs
sachpatel: 👍 in terms of debugging or rewriting, don't worry about it. @mgwilliam wrote it originally and he is good at Congo queries. In terms of performance against Cosmos, will have to test this but I am confident we will be hitting the db less number of times.

src/Jobs/Recruit.Vacancies.Jobs/SpikeJob/SpikeJob.cs
sachpatel: Yeah, I think it will be useful since I seem to create it so many times for debugging purposes.

PR #465, opened 2019-02-12
Gds ux style review

sdholder:
Style changes as result of journey walk through with UX

### DIFF
--- a/src/Employer/Employer.Web/Views/Part1/SearchResultPreview/SearchResultPreview.cshtml
+++ b/src/Employer/Employer.Web/Views/Part1/SearchResultPreview/SearchResultPreview.cshtml
@@ -1,14 +1,15 @@
 ﻿@model Esfa.Recruit.Employer.Web.ViewModels.Part1.SearchResultPreview.SearchResultPreviewViewModel
 <div class="govuk-grid-row">
-    <div class="govuk-grid-column-full">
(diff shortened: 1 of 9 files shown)


src/Employer/Employer.Web/Views/Part1/SearchResultPreview/SearchResultPreview.cshtml
```diff
@@ -1,14 +1,15 @@
...
+            <h2 class="govuk-heading-m govuk-!-margin-bottom-1 govuk-!-margin-top-4">@Model.Title</h2>
             <ul class="govuk-list">
-                <li >@Model.EmployerName</li>
-
-                <li class="govuk-body-s">
-                    <span id="date-and-positions">@Model.NumberOfPositionsCaption</span>
+                <li class="govuk-caption-m">
+                    @Model.EmployerName
+                </li>
+                <li class="govuk-caption-m govuk-!-font-size-16">
```
sachpatel: Could be controversial, is the original `id` in the span element that has been removed used by @sultansfa1  automated tests. If so, we will need to put a `esfa-automation` attribute on the `li` maybe?

sdholder: I've checked with Sultan and he's not using that id so I think it's safe to remove it

PR #477, opened 2019-02-20
Changed QaRulesConfiguration to Singleton

chrisjensenuk:
We're retrieving the QaRules from the DB on every page request for Employer and Provider. So suggesting we change to Singleton.

### DIFF
--- a/src/Shared/Recruit.Vacancies.Client/Ioc/ServiceCollectionExtensions.cs
+++ b/src/Shared/Recruit.Vacancies.Client/Ioc/ServiceCollectionExtensions.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using Esfa.Recruit.Vacancies.Client.Application.Aspects;
+﻿using Esfa.Recruit.Vacancies.Client.Application.Aspects;
 using Esfa.Recruit.Vacancies.Client.Application.Cache;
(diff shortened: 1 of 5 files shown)


src/Shared/Recruit.Vacancies.Client/Ioc/ServiceCollectionExtensions.cs
```diff
@@ -61,7 +61,7 @@ public static void AddRecruitStorageClient(this IServiceCollection services, ICo
                 .Configure<AccountApiConfiguration>(configuration.GetSection("AccountApiConfiguration"))
```
shomavg: Can the namespace be corrected to the location of the file?

shomavg: Also remove unwanted namespace?

shomavg: Change the namespace to point to the correct location.

shomavg: Remove unused namespace please.

chrisjensenuk: 👍

PR #481, opened 2019-02-22
 Disabing MongoCollectionCheck Because of Kestrel Socket Errors

sachpatel:
We get these a lot on AT during deployment and have started to see them more often on PROD.

### DIFF
--- a/src/Employer/Employer.Web/Program.cs
+++ b/src/Employer/Employer.Web/Program.cs
@@ -16,11 +16,7 @@ namespace Esfa.Recruit.Employer.Web
             try
             {
                 logger.Info("Starting up host");
-                var host = CreateWebHostBuilder(args).Build();
-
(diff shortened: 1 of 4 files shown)


src/Employer/Employer.Web/Program.cs
```diff
@@ -16,11 +16,7 @@ public static void Main(string[] args)
             try
             {
                 logger.Info("Starting up host");
-                var host = CreateWebHostBuilder(args).Build();
-
-                CheckInfrastructure(host.Services, logger);
-
```
shomavg: The CheckInfrastructure() also creates the tablestorage. Do we not want this?

sachpatel: I am not merging this, we don't have tablestorage turned on in AT. I am just trying to find what the issue is. I need a branch to deploy to AT and the best way for it to create the initial release is to create a PR.

PR #482, opened 2019-02-25
ER-972 provider navigation menu

sachpatel:
This is basic.

Going forward we will be using a shared partial across DAS for the employer and Provider menu's these should also take not of the whitelisting and env differences for showing particular menu items.

I have copied in the way that the footer is on the PAS website on test, including a survey link to AccessMyLevy, we can change this as we get a survey for provider recruit or drop it all together.

I have left out the reservations link which might appear on the other env's because it is not showing up on TEST yet. We will have to revisit this when going through the env's to make sure the menu items are the same across other das services.

### DIFF
--- a/src/Provider/Provider.Web/Controllers/ExternalLinksController.cs
+++ b/src/Provider/Provider.Web/Controllers/ExternalLinksController.cs
@@ -1,0 +1,75 @@
+﻿using Esfa.Recruit.Provider.Web.Configuration;
+using Esfa.Recruit.Provider.Web.Configuration.Routing;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Esfa.Recruit.Provider.Web.Controllers
+{
+    [Route(RoutePaths.AccountRoutePath)]
+	public class ExternalLinksController : Controller
+    {
+        private readonly ProviderApprenticeshipsLinkHelper _linkHelper;
+
+        public ExternalLinksController(ProviderApprenticeshipsLinkHelper linkHelper)
+        {
+            _linkHelper = linkHelper;
+        }
+
+        [HttpGet("account-home", Name = RouteNames.Dashboard_Account_Home)]
+        public IActionResult AccountHome()
+        {
+            return Redirect(_linkHelper.AccountHome);
+        }
+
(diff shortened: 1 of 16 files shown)


src/Provider/Provider.Web/Controllers/ExternalLinksController.cs
```diff
@@ -0,0 +1,74 @@
...
+	public class ExternalLinksController : Controller
+    {
+        private readonly ProviderApprenticeshipsLinkHelper _linkHelper;
+
+        public ExternalLinksController(ProviderApprenticeshipsLinkHelper linkHelper)
+        {
+            _linkHelper = linkHelper;
+        }
+
+        [HttpGet(RoutePaths.AccountRoutePath + "/account-home", Name = RouteNames.Dashboard_Account_Home)]
```
chrisjensenuk: Can we put `RoutePaths.AccountRoutePath` as a class level attribute?

src/Shared/Recruit.Vacancies.Client/Infrastructure/Mongo/MongoDbCollectionBase.cs
```diff
@@ -39,8 +39,9 @@ protected IMongoCollection<T> GetCollection<T>()
         {
             var settings = MongoClientSettings.FromUrl(new MongoUrl(_config.ConnectionString));
             settings.SslSettings = new SslSettings { EnabledSslProtocols = SslProtocols.Tls12 };
+            settings.ConnectionMode = ConnectionMode.Direct;
```
chrisjensenuk: What does this do?

src/Provider/Provider.Web/Controllers/ExternalLinksController.cs
sachpatel: 👍

src/Shared/Recruit.Vacancies.Client/Infrastructure/Mongo/MongoDbCollectionBase.cs
sachpatel: 😱 mistake, will remove.

PR #484, opened 2019-03-01
Create reference number in-process when cloning

dev4valtech:
I wanted to do this refactoring outside of provider clone story, hence this PR so people can get an idea and debate on this.  

I have changed the event being raised from CloneVacancyCommandHandler to VacancyCreatedEvent instead of VacancyClonedEvent. I think though the command is cloning a vacancy, effectively we are creating a new one and hence VacancyCreatedEvent made more sense to me. This has made VacancyClonedEvent redundant and hence I have removed it.

Thoughts are welcome.


mgwilliam:
@dev4valtech i'd prefer the call to assign the number to be done within the vacancy client rather than from the orchestrator - reason being it's an integral part of the clone feature which we'd need to occur when cloning via the v2 API (for example).

PR #491, opened 2019-03-08
ER-470 Reporting command

chrisjensenuk:
Next batch of changes.  This is to do with the create report command.

Fluent validation defined in the various web projects now need to be explicitly registered in IOC as I have removed the RegisterValidatorsFromAssemblyContaining function call.

- Created Reports collection
- Added CreateReportCommand
- Added CreateReportCommandHandler
- Adds a report to the Reports collection


src/Provider/Provider.Web/ViewModels/Reports/ProviderApplicationsReport/ProviderApplicationsReportCreateEditModel.cs
```diff
@@ -1,7 +1,12 @@
-﻿namespace Esfa.Recruit.Provider.Web.ViewModels.Reports.ProviderApplicationsReport
+﻿using Microsoft.AspNetCore.Mvc;
+
+namespace Esfa.Recruit.Provider.Web.ViewModels.Reports.ProviderApplicationsReport
 {
     public class ProviderApplicationsReportCreateEditModel
     {
+        [FromRoute]
```
dev4valtech: This should be derived from User claims

src/Shared/Recruit.Vacancies.Client/Domain/Events/ReportCreatedEvent.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using Esfa.Recruit.Vacancies.Client.Domain.Messaging;
+using MediatR;
+
+namespace Esfa.Recruit.Vacancies.Client.Domain.Events
+{
+    public class ReportCreatedEvent : EventBase, INotification
```
dev4valtech: What is the purpose of this event?

src/Employer/Employer.Web/Configuration/ConfigurationExtensions.cs
```diff
@@ -88,8 +88,7 @@ public static void AddMvcService(this IServiceCollection services, IHostingEnvir
                 opts.Filters.AddService<PlannedOutageResultFilter>();
 
                 opts.AddTrimModelBinderProvider(loggerFactory);
-            })
-            .AddFluentValidation(fv =>fv.RegisterValidatorsFromAssemblyContaining<Startup>())  
+            })
```
shomavg: Can you please test if the validations are still working on application reviews?

src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/CreateReportCommandHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        private readonly ILogger<CreateReportCommandHandler> _logger;
+        private readonly IReportRepository _repository;
+        private readonly IMessaging _messaging;
+        private readonly ITimeProvider _timeProvider;
+
+        public CreateReportCommandHandler(
+            ILogger<CreateReportCommandHandler> logger,
+            IReportRepository repository, 
+            IMessaging messaging,
```
shomavg: In there for future use?

src/Shared/Recruit.Vacancies.Client/Domain/Events/ReportCreatedEvent.cs
chrisjensenuk: 👍 leftovers. Will remove

src/Provider/Provider.Web/ViewModels/Reports/ProviderApplicationsReport/ProviderApplicationsReportCreateEditModel.cs
chrisjensenuk: I'm happy with this. In future we could have a god user that can navigate between providers so we should let the route dictate which provider the user is targeting. So I think decoupling resource identification from authorisation is a good thing.

The authorisation filter is triggered on every page request so I know I can trust this ukprn.

src/Employer/Employer.Web/Configuration/ConfigurationExtensions.cs
chrisjensenuk: 👍 Will do

chrisjensenuk: Doh! Still have to configure FluentValidation for MVC. Have now fixed. Good shout!

src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/CreateReportCommandHandler.cs
chrisjensenuk: That was the plan but in the end I'm not using messaging so I have now removed.

src/Provider/Provider.Web/ViewModels/Reports/ProviderApplicationsReport/ProviderApplicationsReportCreateEditModel.cs
dev4valtech: While I am not sure what the "god" user requirements are, I am happy that authorisation filter is safe guarding the value in URL.

PR #492, opened 2019-03-08
ER-470 reporting messaging

chrisjensenuk:
Next batch of changes to do with messaging.

Once the report has been created and stored in the repository a message is added to the `report-queue`.  A Job listens to messages posted to this queue and once triggered calls a service to generate the report.

- I have created an abstract `StorageQueueBase` class to share some common functionality which is utilised by the existing `StorageQueueEventStore` and the new `StorageQueueReportQueue`.


chrisjensenuk:
Hey @dev4valtech Not an a$$, a reasonable question. The easiest answer is that its what is in the spec! https://skillsfundingagency.atlassian.net/wiki/spaces/RAAV2/pages/1133838361/Recruitment+reports

@mgwilliam Is this how you envisaged the queuing of reports working?

mgwilliam:
@chrisjensenuk I've not reviewed the code in this PR in detail but my understanding is that the queue is used to store a request to create a report. for now those requests will come directly from the users clicking in the front end to create a report.

in future sprints we'll possibly/likely introduce scheduled reports (eg. a monthly summary of all vacancy statuses) so these won't be triggered by a user directly. instead, we'll likely have a monthly task to enqueue the scheduled report for all providers / employers.

@dev4valtech - it's a fair question and we certainly **could have** implemented it using the existing domain events, etc. for the events we have the abstraction "event store" which happens to be implemented as a queue (currently). the v2 POC also used a cosmos collection for the "event store" which we may well revert to as it gave us a persistent record of events and we could use the cosmos trigger in the jobs.

as such, "event store" needs to remain abstract. ie. we shouldn't be tying it directly to a message queue as that may well change.

on the other hand, the reports wouldn't ever be events as they don't represent a change in the state of the domain. instead, they're an instruction to generate a report (ie. no domain state change)

so IMO there's a clear distinction between the events and the reports.

give me a shout if this isn't clear or doesn't make sense!

PR #493, opened 2019-03-08
ER-968 Implements migration script for vacancyReviews

dev4valtech:
I have tested the migration script with 2000 documents using the emulator. Batch process with insertmany failed on requests, this one just breezed through.


src/Data/Recruit.ChangeScripts/050-vacancyReviews_migration.js
```diff
@@ -0,0 +1,7 @@
+{
+    print((new Date()) + "Start migrating documents from reviews to vacancyReviews collection.");
```
sachpatel: You need a space before `Start` and same on `Finish` line.

Could you also print the document count on source collection before migration and after.

src/Data/Recruit.ChangeScripts/050-vacancyReviews_migration.js
```diff
@@ -0,0 +1,7 @@
...
+
+    db.reviews.find().forEach( function(x){db.vacancyReviews.insert(x);} );
```
sachpatel: What is to stop the `request rate` error, do you need a sleep?

dev4valtech: I have tried it on emulator and it worked without any errors, so I am hoping it would work. Note that InsertMany failed locally so I am assuming that local run is behaving similar to hosted.

dev4valtech:
@sachpatel There is no option but to have a little down time. I was not sure where the script would go. Happy to move it elsewhere.

chrisjensenuk:
We also need to point to the new collection so there is a code change to go along with this so can see the argument for doing as part of deployment.

dev4valtech:
@chrisjensenuk on your idempotent comment, since we are inserting using the id, it should ideally fail to insert again, so I am relying on Cosmos to fail it. Trying to keep this as simple as possible without adding any unnecessary complication.

chrisjensenuk:
@dev4valtech I agree. But if `db.vacancyReviews.insert(x);` fails doesn't the whole job fail rather than just skipping that particular document?

dev4valtech:
@chrisjensenuk, I have tested it locally, it carries on error.

src/Data/Recruit.ChangeScripts/050-vacancyReviews_migration.js
sachpatel: all you need is the change script without a sequence number prefix, it won't get run as part of deployment so I suggest using that as we will run this script hopefully just once and not part of deployment. After we have run the script then we can deploy the version of the code with the `MongoCollectionNames.cs` update.

PR #501, opened 2019-03-13
Add applications columns to dashboards

sachpatel:
This is something I missed out of the provider allow create applications vacancy story. Got kicked back from PO acceptance. An addition has been made to Employer as well.

### DIFF
--- a/src/Shared/Recruit.Shared.Web/ViewModels/VacancySummaryViewModel.cs
+++ b/src/Shared/Recruit.Shared.Web/ViewModels/VacancySummaryViewModel.cs
@@ -23,7 +23,7 @@ namespace Esfa.Recruit.Shared.Web.ViewModels
 
         public bool HasVacancyReference => VacancyReference.HasValue;
         public bool HasNoVacancyReference => !HasVacancyReference;
-        public bool IsApplicationsVacancy => (Status== VacancyStatus.Live || Status == VacancyStatus.Closed) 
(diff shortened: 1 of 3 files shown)


src/Shared/Recruit.Shared.Web/ViewModels/VacancySummaryViewModel.cs
```diff
@@ -23,7 +23,7 @@ public class VacancySummaryViewModel
 
         public bool HasVacancyReference => VacancyReference.HasValue;
         public bool HasNoVacancyReference => !HasVacancyReference;
-        public bool IsApplicationsVacancy => (Status== VacancyStatus.Live || Status ==VacancyStatus.Closed) 
+        public bool IsApplicationsVacancy => (Status== VacancyStatus.Live || Status == VacancyStatus.Closed)
```
dev4valtech: CanShowApplicationsOnVacancy instead of IsApplicationsVacancy?

sachpatel: I don't agree with your suggestions. I am calling it `CanShowVacancyApplicationsCount`

dev4valtech: If that's how you want to be 😉  ... sounds good 👍

PR #502, opened 2019-03-15
ER-1012_Confirmation for application status in provider.

shomavg:
Mostly copy paste from employer.


src/Employer/Employer.Web/Controllers/ApplicationReviewController.cs
```diff
@@ -15,6 +15,7 @@ namespace Esfa.Recruit.Employer.Web.Controllers
     public class ApplicationReviewController : Controller
     {
         private readonly ApplicationReviewOrchestrator _orchestrator;
+        private const string TempDateARModel = "ApplicationReviewEditModel";
```
chrisjensenuk: Typo TempData

shomavg: fixed the typo.

PR #503, opened 2019-03-18
ER-1022_Text change as per template.

shomavg:
Minor text and style changes on the _reviewHeader page.


shomavg:
moved to a class.

PR #507, opened 2019-03-19
ER-470 reporting service

chrisjensenuk:
This PR is the generating of the report and the downloading as a CSV.

The ReportService uses a different strategy for each report type. We only have the 1 report at the moment. The strategy queries Cosmos and returns a json string containing the report data.

The report data is then saved against the Report entity and the Report status is set to 'Generated'.

From the Provider web site a user clicks to download a report. A CSV is generated at runtime using the Report entities data and is downloaded.

- I've purposely left the mongo query as a string to make debugging easier.
- I've not attempted to create Entities to desearialise the reports into. I've tried to keep the reporting service as stand alone and generic as possible so it just converts bson to json data and json to csv directly.


src/Provider/Provider.Web/Orchestrators/Reports/ProviderApplicationsReportOrchestrator.cs
```diff
@@ -64,7 +65,9 @@ public Task<Guid> PostCreateViewModelAsync(ProviderApplicationsReportCreateEditM
                     throw new Exception($"Cannot handle this date range type:{model.DateRange.ToString()}");
             }
 
-            return _client.CreateProviderApplicationsReportAsync(model.Ukprn, fromDate, toDate, user);
+            var reportName = $"{fromDate.AsGdsDate()} to {toDate.AsGdsDate()}";
```
dev4valtech: Since we are anticipating other reports, perhaps we should have a way to identify the type of report requested by appending to the name or some other means.

chrisjensenuk: The viewing of multiple report types from the dashboard has not been considered yet so this will need to be revisited at this time so I've purposefully not been too clever here as I don't want to presume too much.

chrisjensenuk: As discussed just now. We know what 'type' the report is so if we need to differentiate between report types in future we will have enough information to do this and not need to patch the reports.

PR #508, opened 2019-03-20
ER-470 Reporting delete job

chrisjensenuk:
This is the last PR for this ticket! 
This is a nightly job that deletes reports that are over 7 days old.

### DIFF
--- a/src/Shared/Recruit.Vacancies.Client/Infrastructure/Repositories/MongoDbReportRepository.cs
+++ b/src/Shared/Recruit.Vacancies.Client/Infrastructure/Repositories/MongoDbReportRepository.cs
@@ -66,5 +66,17 @@ namespace Esfa.Recruit.Vacancies.Client.Infrastructure.Repositories
 
             return result;
         }
+
+        public async Task<int> DeleteReportsCreatedBeforeAsync(DateTime requestedOn)
+        {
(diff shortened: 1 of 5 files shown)


src/Shared/Recruit.Vacancies.Client/Infrastructure/Repositories/MongoDbReportRepository.cs
```diff
@@ -66,5 +66,17 @@ public async Task<Report> GetReportAsync(Guid reportId)
...
             return result;
         }
+
+        public async Task<int> DeleteReportsCreatedBeforeAsync(DateTime requestedOn)
+        {
+            var filter = Builders<Report>.Filter.Lt(r => r.RequestedOn, requestedOn);
+            var collection = GetCollection<Report>();
+
+            var result = await RetryPolicy.ExecuteAsync(async _ =>
+                    await collection.DeleteManyAsync(filter),
```
dev4valtech: Delete Many can run into errors if too many documents are aggregated for deletion. May be we can find a limit that works and batch the process. Or is the strategy is to wait till we encounter errors?

chrisjensenuk: Good shout. I'm going to leave this for the moment as not to over optimise too soon.  We are a while away yet before this might cause problems and its non critical if it does.

PR #512, opened 2019-03-25
ER-1139 do not allow whitespace form entry strings

sachpatel:


setup/containers/docker-compose.yml
```diff
@@ -88,7 +88,7 @@ services:
     image: mcr.microsoft.com/azurestorage-azurite:2.6.7
     container_name: azurite
     volumes:
-      - azurite1:/opt/azurite/data
+      - azurite1:/opt/azurite/folder
```
chrisjensenuk: Why has this changed?

sachpatel: QueryStore table storage was empty every time I started the container. Maybe the newer image changed the directory used.

PR #521, opened 2019-04-01
ER-866 whitelisting providers

shomavg:


shomavg:
still need some testing.

PR #522, opened 2019-04-01
ER-866_whitelisting_providers

shomavg:
Whitelisting providers.

### DIFF
--- a/src/Employer/Esfa.Recruit.Azure.Resources/azuredeploy.json
+++ b/src/Employer/Esfa.Recruit.Azure.Resources/azuredeploy.json
@@ -635,6 +635,10 @@
 			  {
                 "name": "FaaConfiguration:StorageConnectionString",
                 "value": "[parameters('FaaConfiguration').StorageConnectionString]"
+              },
+			  {
+                "name": "WhitelistedProvidersList",
+                "value": "[parameters('WhitelistedProvidersList')]"
               }
             ]
           },
--- a/src/Employer/Esfa.Recruit.Azure.Resources/azuredeploy.json
+++ b/src/Employer/Esfa.Recruit.Azure.Resources/azuredeploy.json
@@ -110,6 +110,10 @@
       "type": "bool",
       "defaultValue": false
     },
+	"WhitelistedProvidersList":{
+	  "type": "string",
+      "defaultValue": ""
+	},
     "SharedAppServicePlanName": {
       "type": "string"
     },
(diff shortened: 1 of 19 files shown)


src/Provider/Provider.Web/Configuration/ConfigurationExtensions.cs
```diff
@@ -71,10 +72,14 @@ public static void AddMvcService(this IServiceCollection services, IHostingEnvir
                 opts.Filters.AddService<PlannedOutageResultFilter>();
 
                 opts.AddTrimModelBinderProvider(loggerFactory);
+                if (EnvironmentNames.GetTestEnvironmentNames().Contains(hostingEnvironment.EnvironmentName.ToUpper()) == false)
```
dev4valtech: The if condition can be converted into meaningful function with name like "IsProductionEquivalentEnvironment"

src/Employer/Esfa.Recruit.Azure.Resources/azuredeploy.json
```diff
@@ -110,6 +110,10 @@
       "type": "bool",
       "defaultValue": false
     },
+	"WhitelistedProvidersList":{
```
dev4valtech: This needs to be added in the parameters file as well.

shomavg: ok..will do!

src/Provider/Provider.Web/Configuration/ConfigurationExtensions.cs
shomavg: It's the same functionality used for Employer... ! i will give it a meaningful name.

src/Provider/Provider.Web/Controllers/ErrorController.cs
```diff
@@ -111,6 +116,15 @@ private IActionResult PageNotFound()
         {
             Response.StatusCode = (int)HttpStatusCode.NotFound;
             return View(ViewNames.PageNotFound);
+        }       
+
+        // Blocked provider url for private beta
```
chrisjensenuk: We don't need this comment

src/Employer/Esfa.Recruit.Azure.Resources/azuredeploy.json
sachpatel: Where is this even used? (rhetorical question)

shomavg: We provide a list of UKPRN's in the json file.

PR #524, opened 2019-04-02
Show deleted message on QA dashboard. Not building.

shomavg:
Pre-Review code.

### DIFF
--- a/src/QA/QA.Web/Orchestrators/DashboardOrchestrator.cs
+++ b/src/QA/QA.Web/Orchestrators/DashboardOrchestrator.cs
@@ -38,7 +38,7 @@ namespace Esfa.Recruit.Qa.Web.Orchestrators
             {
                 var inProgressSummaries = await _vacancyClient.GetVacancyReviewsInProgressAsync();
 
-                var inProgressVacanciesVmTasks = inProgressSummaries.Select(async v => await MapToViewModelAsync(v, vacancyUser)).ToList();
+                var inProgressVacanciesVmTasks = inProgressSummaries.Select(v => MapToViewModelAsync(v, vacancyUser)).ToList();
                 var inProgressVacanciesVm = await Task.WhenAll(inProgressVacanciesVmTasks);
                 vm.InProgressVacancies = inProgressVacanciesVm.ToList();
             }
@@ -47,7 +47,6 @@ namespace Esfa.Recruit.Qa.Web.Orchestrators
 
             vm.LastSearchTerm = searchTerm;
             var searchResults = await _vacancyClient.GetSearchResultsAsync(searchTerm);
-
             var searchResultsVmTasks = searchResults.Select(v => MapToViewModelAsync(v, vacancyUser)).ToList();
             var searchResultsVm = await Task.WhenAll(searchResultsVmTasks);
             vm.SearchResults = searchResultsVm.ToList();
--- a/src/QA/QA.Web/Orchestrators/DashboardOrchestrator.cs
+++ b/src/QA/QA.Web/Orchestrators/DashboardOrchestrator.cs
@@ -48,10 +48,12 @@ namespace Esfa.Recruit.Qa.Web.Orchestrators
             return vm;
         }
 
-        private VacancyReviewSearchResultViewModel MapToViewModel(VacancyReview vacancyReview, VacancyUser vacancyUser)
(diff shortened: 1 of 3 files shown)


src/QA/QA.Web/Views/Dashboard/Index.cshtml
```diff
@@ -119,6 +119,10 @@
                     <td class="govuk-table__cell">
                         <span asp-show="@vacancy.ShowAssignmentInfoCaption">@assignmentInfoCaption<br /></span>
                         <a asp-show="@vacancy.CanShowReviewLink" asp-route="@RouteNames.Vacancy_Review_Get" asp-route-reviewId="@vacancy.ReviewId" class="govuk-link">Review</a>
+                        @if (vacancy.IsVacancyDeleted)
+                        {
+                            <p>This vacancy has been deleted</p>
```
chrisjensenuk: Can you use the tag helper instead of the `if` statement `<p asp-show="@vacancy.IsVacancyDeleted">`

src/QA/QA.Web/Orchestrators/DashboardOrchestrator.cs
```diff
@@ -65,7 +73,8 @@ private VacancyReviewSearchResultViewModel MapToViewModel(VacancyReview vacancyR
                 ReviewId = vacancyReview.Id,
                 IsClosed =  vacancyReview.Status == ReviewStatus.Closed,
                 SubmittedDate = vacancyReview.VacancySnapshot.SubmittedDate.GetValueOrDefault(),
-                IsAvailableForReview = isAvailableForReview
+                IsAvailableForReview = isAvailableForReview,
+                IsVacancyDeleted=vacancy.IsDeleted
```
chrisjensenuk: spacing

src/QA/QA.Web/Views/Dashboard/Index.cshtml
```diff
@@ -119,7 +119,8 @@
                     <td class="govuk-table__cell">
                         <span asp-show="@vacancy.ShowAssignmentInfoCaption">@assignmentInfoCaption<br /></span>
                         <a asp-show="@vacancy.CanShowReviewLink" asp-route="@RouteNames.Vacancy_Review_Get" asp-route-reviewId="@vacancy.ReviewId" class="govuk-link">Review</a>
-                        <span asp-show="@vacancy.IsClosed">@vacancy.ClosedReviewAssignmentCaption</span>
+                        <p asp-show="@vacancy.IsVacancyDeleted">This vacancy has been deleted.</p>
+                        <span asp-show="@vacancy.IsClosed && !@vacancy.IsVacancyDeleted">@vacancy.ClosedReviewAssignmentCaption</span>
```
chrisjensenuk: We put helper properties in the VM to prevent these kind of expressions from complicating the view.  Do we care that the vacancy is not deleted?  I think it might be useful to know who reviewed the vacancy before it got deleted.  As it stands at the moment you can't delete a closed vacancy anyway so you'll never have a deleted closed vacancy so this check here is superfluous.  If this is indeed the logic can you put it into a helper bool property in the VM.

src/QA/QA.Web/Orchestrators/DashboardOrchestrator.cs
```diff
@@ -37,22 +37,30 @@ public async Task<DashboardViewModel> GetDashboardViewModelAsync(string searchTe
             if (vm.DisplayInProgressVacancies)
             {
                 var inProgressSummaries = await _vacancyClient.GetVacancyReviewsInProgressAsync();
-                vm.InProgressVacancies = inProgressSummaries.Select(v => MapToViewModel(v, vacancyUser)).ToList();
+
+                var inProgressVacanciesVmTasks = inProgressSummaries.Select(async v => await MapToViewModelAsync(v, vacancyUser)).ToList();
```
chrisjensenuk: We don't need the async/await here as we're not doing anything with the result inside the lambda. So we can change it so it looks like the one on line 51.

src/QA/QA.Web/Views/Dashboard/Index.cshtml
shomavg: I did speak to Jithesh and Sultan about it. They are very clear and want the message 'This vacancy is deleted' only. They do not want the text that shows who reviewed it.

src/QA/QA.Web/Orchestrators/DashboardOrchestrator.cs
shomavg: Made the changes.

PR #533, opened 2019-04-08
ER-1172 Changes empname to legalentityname in dashboard projections

dev4valtech:


src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/VacancySummariesProvider/VacancySummaryMapper.cs
```diff
@@ -13,7 +13,7 @@ internal static VacancySummary MapFromVacancySummaryAggQueryResponseDto(VacancyS
                 Id = vacSummaryDetail.VacancyGuid,
                 Title = vacSummaryDetail.Title,
                 VacancyReference = vacSummaryDetail.VacancyReference,
-                EmployerName = vacSummaryDetail.EmployerName,
+                EmployerName = vacSummaryDetail.LegalEntityName,
```
chrisjensenuk: Can we change EmployerName to LegalEntityName too?

dev4valtech: I am not sure where this is being used and what the impact , hence left it. so question is back on you, should we do that?

chrisjensenuk: In an ideal world we should update this too however there would be data fixes required to update the dashboard. So taking into account current time pressures this is the least risk thing to do so this is OK.

PR #534, opened 2019-04-08
Fixing wayward angle bracket

chrisjensenuk:


chrisjensenuk:
@sachpatel It's included

PR #537, opened 2019-04-09
ER-1179 Fixes trading name in readonly view

dev4valtech:


src/Shared/UnitTests/Vacancies.Client/Application/Services/EmployerNameServiceTest.cs
```diff
@@ -0,0 +1,96 @@
...
+
+            var result = sut.GetEmployerNameAsync(Guid.NewGuid()).Result;
+
+            result.Should().Be(legalEntityName);
+        }
+
+        [Theory]
+        [InlineData(VacancyStatus.Draft)]
+        [InlineData(VacancyStatus.Referred)]
+        public void ShouldReturnTradingName(VacancyStatus status)
```
sachpatel: async Task?

src/Shared/UnitTests/Vacancies.Client/Application/Services/EmployerNameServiceTest.cs
```diff
@@ -0,0 +1,96 @@
...
+
+            var result = sut.GetEmployerNameAsync(Guid.NewGuid()).Result;
+
+            result.Should().Be(employerName);
+        }
+
+        [Theory]
+        [InlineData(VacancyStatus.Draft)]
+        [InlineData(VacancyStatus.Referred)]
+        public void ShouldReturnVacancyLegalEntityName(VacancyStatus status)
```
sachpatel: async Task?

src/Shared/UnitTests/Vacancies.Client/Application/Services/EmployerNameServiceTest.cs
dev4valtech: thought about it and went ahead with no tasks deliberately as there is no need for it.

src/Shared/UnitTests/Vacancies.Client/Application/Services/EmployerNameServiceTest.cs
dev4valtech: thought about it and went ahead with no tasks deliberately as there is no need for it.

PR #551, opened 2019-04-13
ER-1192 - Provider service claim.

chrisjensenuk:
Funny what you wake up thinking about....

I'd removed this `else` in the previous PR which added the Provider service claim authorisation.  

If UKPRN isn't specified in the route then we should allow.


chrisjensenuk:
@sachpatel 👍

PR #556, opened 2019-04-18
ER-1191 text change

shomavg:


src/Provider/Provider.Web/Orchestrators/DashboardOrchestrator.cs
```diff
@@ -99,43 +99,40 @@ private int SanitizePage(int page, int totalVacancies)
...
 
-            return new SelectListItem($"{text} ({count})", value, optionStatus == filterStatus );
+            return new SelectListItem($"{text}", value, optionStatus == filterStatus );
         }
 
         private string GetFilterHeading(int totalVacancies, VacancyStatus? filterStatus)
         {
             if (totalVacancies == 1 && filterStatus.HasValue == false)
-                return "Showing 1 vacancy";
+                return "Show 1 vacancy";
```
sachpatel: if it is just 1 then should it not be `Show vacancy` only

src/Provider/Provider.Web/Orchestrators/DashboardOrchestrator.cs
```diff
@@ -99,43 +99,40 @@ private int SanitizePage(int page, int totalVacancies)
...
             {
-                GetFilterSelectListItem(vacancies, null, filterStatus),
-                GetFilterSelectListItem(vacancies, VacancyStatus.Draft, filterStatus),
-                GetFilterSelectListItem(vacancies, VacancyStatus.Submitted, filterStatus),
-                GetFilterSelectListItem(vacancies, VacancyStatus.Live, filterStatus),
-                GetFilterSelectListItem(vacancies, VacancyStatus.Closed, filterStatus),
-                GetFilterSelectListItem(vacancies, VacancyStatus.Referred, filterStatus),
+                GetFilterSelectListItem(null, filterStatus),
+                GetFilterSelectListItem(VacancyStatus.Live, filterStatus),
+                GetFilterSelectListItem(VacancyStatus.Referred, filterStatus),
```
sachpatel: Why do we have a SelectListItem in the orchestrator, that is a View concern

shomavg: This is build on the existing filter functionality done by Chris. I have only modified the text as per the ticket.

chrisjensenuk: @sachpatel Orchestrators are ViewModel aware I used `SelectListItem` as I didn't see the point in create our own VM to do this.

src/Provider/Provider.Web/Orchestrators/DashboardOrchestrator.cs
```diff
@@ -65,7 +65,7 @@ public async Task<DashboardViewModel> GetDashboardViewModelAsync(long ukprn, str
                 Pager = pager,
                 IsFiltered = filterStatus.HasValue,
                 ShowFilter = vacancies.Select(v => v.Status).Distinct().Count() > 1,
```
chrisjensenuk: We can remove this bit as we always want to show it now.

src/Provider/Provider.Web/Orchestrators/DashboardOrchestrator.cs
sachpatel: ViewModel aware yes but we haven't been mixing MVC into Orchestrators previously or HTTP 😬 . Having a look this is going to be quite a bit to untangle so I'll leave it for you guys to manage with the changes being made to the dashboard.

chrisjensenuk: @sachpatel  No untangling to be done. We can just create our own class with `Text`, `Value` & `Selected` properties.

Thinking about it - the drop down is now completely static so we don't need to do any of this anymore we can just put it in the view.  We had to construct the list server-side previously so we could include the counts.  @shomavg Can you kill all this and just add the <options> in the HTML?

src/Provider/Provider.Web/Orchestrators/DashboardOrchestrator.cs
```diff
@@ -88,54 +121,36 @@ private async Task<List<VacancySummary>> GetVacanciesAsync(long ukprn)
 
         private int SanitizePage(int page, int totalVacancies)
         {
-            return (page < 0 || page > (int)Math.Ceiling((double)totalVacancies / VacanciesPerPage)) ? 1 : page;
+            return page < 0 || page > (int)Math.Ceiling((double)totalVacancies / VacanciesPerPage) ? 1 : page;
         }
 
-        private VacancyStatus? SanitizeFilter(string filter)
+        private FilteringOptions? SanitizeFilter(string filter)
```
chrisjensenuk: I don't think this function should return a nullable now we have `FilteringOptions`
Instead of returning null, shouldn't this return `FilteringOptions.All`.

src/Provider/Provider.Web/Orchestrators/DashboardOrchestrator.cs
```diff
@@ -63,16 +59,53 @@ public async Task<DashboardViewModel> GetDashboardViewModelAsync(long ukprn, str
...
+                IsFiltered = filterStatus.HasValue,                 
+                Filter = filterStatus?.ToString(),
                 ResultsHeading = GetFilterHeading(filteredVacanciesTotal, filterStatus),
                 HasVacancies = vacancies.Any()
             };
 
             return vm;
         }
 
+        private List<VacancySummary> GetFilteredVacancies(List<VacancySummary> vacancies, FilteringOptions? filterStatus)
```
chrisjensenuk: `filterStatus` shouldn't be nullable as we now have `FilteringOptions.All`

src/Provider/Provider.Web/Orchestrators/DashboardOrchestrator.cs
```diff
@@ -63,16 +59,53 @@ public async Task<DashboardViewModel> GetDashboardViewModelAsync(long ukprn, str
             {
                 Vacancies = vacanciesVm,
                 Pager = pager,
-                IsFiltered = filterStatus.HasValue,
-                ShowFilter = vacancies.Select(v => v.Status).Distinct().Count() > 1,
-                FilterOptions = GetFilterSelectOptions(vacancies, filterStatus),
+                IsFiltered = filterStatus.HasValue,
```
chrisjensenuk: We shouldn't need to set this.  We can have a property on VM `IsFiltered => Filter == FilterOptions.All`

src/Provider/Provider.Web/Orchestrators/DashboardOrchestrator.cs
```diff
@@ -63,16 +59,53 @@ public async Task<DashboardViewModel> GetDashboardViewModelAsync(long ukprn, str
...
+                    filteredVacancies = vacancies.Where(v => v.NoOfNewApplications > 0);
+                    break;
+                case FilteringOptions.AllApplications:
+                    filteredVacancies = vacancies.Where(v =>
+                        v.NoOfSuccessfulApplications > 0 || v.NoOfUnsuccessfulApplications > 0 ||
+                        v.NoOfNewApplications > 0);
+                    break;
+                case FilteringOptions.ClosingSoon:
+                    filteredVacancies = vacancies.Where(v =>
+                        v.ClosingDate <= DateTime.Now.AddDays(5) && v.Status == VacancyStatus.Live);
```
chrisjensenuk: We should never use DateTime.Now use the `ITimeProvider`

src/Provider/Provider.Web/Orchestrators/DashboardOrchestrator.cs
```diff
@@ -63,16 +59,53 @@ public async Task<DashboardViewModel> GetDashboardViewModelAsync(long ukprn, str
...
+                    break;
+                case FilteringOptions.ClosingSoonWithNoApplications:
+                    filteredVacancies = vacancies.Where(v =>
+                        v.ClosingDate <= DateTime.Now.AddDays(5) && v.Status == VacancyStatus.Live && (v.NoOfSuccessfulApplications == 0 || v.NoOfUnsuccessfulApplications == 0 ||
```
chrisjensenuk: Use `ITimeProvider` instead of `Datetime.Now`

src/Provider/Provider.Web/Orchestrators/DashboardOrchestrator.cs
```diff
@@ -88,54 +121,36 @@ private async Task<List<VacancySummary>> GetVacanciesAsync(long ukprn)
 
         private int SanitizePage(int page, int totalVacancies)
         {
-            return (page < 0 || page > (int)Math.Ceiling((double)totalVacancies / VacanciesPerPage)) ? 1 : page;
+            return page < 0 || page > (int)Math.Ceiling((double)totalVacancies / VacanciesPerPage) ? 1 : page;
```
chrisjensenuk: Can you change this back please. Its more readable before.

src/Provider/Provider.Web/Orchestrators/DashboardOrchestrator.cs
```diff
@@ -88,54 +121,36 @@ private async Task<List<VacancySummary>> GetVacanciesAsync(long ukprn)
...
 
-        private VacancyStatus? SanitizeFilter(string filter)
+        private FilteringOptions? SanitizeFilter(string filter)
         {
-            if (Enum.TryParse(typeof(VacancyStatus), filter, out var status))
-                return (VacancyStatus)status;
+            if (Enum.TryParse(typeof(FilteringOptions), filter, out var status))
+                return (FilteringOptions)status;
 
             return null;
```
chrisjensenuk: return `FilteringOptions.All`

src/Provider/Provider.Web/Orchestrators/DashboardOrchestrator.cs
chrisjensenuk: Does this text still need changing as per Sachin's comment?

src/Provider/Provider.Web/Orchestrators/DashboardOrchestrator.cs
```diff
@@ -88,54 +121,36 @@ private async Task<List<VacancySummary>> GetVacanciesAsync(long ukprn)
...
-
-            var value = optionStatus.HasValue ? optionStatus.Value.ToString() : "All";
-
-            var text = optionStatus.HasValue ? optionStatus.Value.GetDisplayName() : "All Vacancies";
-
-            return new SelectListItem($"{text} ({count})", value, optionStatus == filterStatus );
-        }
-
-        private string GetFilterHeading(int totalVacancies, VacancyStatus? filterStatus)
+        private string GetFilterHeading(int totalVacancies, FilteringOptions? filterStatus)
```
chrisjensenuk: `filterStatus` should no longer be nullable

src/Provider/Provider.Web/ViewModels/DashboardViewModel.cs
```diff
@@ -11,14 +11,13 @@ public class DashboardViewModel
         public string WarningMessage { get; internal set; }
         public string InfoMessage { get; internal set; }
         public PagerViewModel Pager { get; internal set; }
-        public IEnumerable<SelectListItem> FilterOptions { get; internal set; }
-        public bool ShowFilter { get; internal set; }
+        public IEnumerable<SelectListItem> FilterOptions { get; internal set; }
```
chrisjensenuk: We don't need this anymore

src/Provider/Provider.Web/Views/Dashboard/Dashboard.cshtml
```diff
@@ -119,7 +132,7 @@
         $(function() {
             $("#filter-submit").hide();
 
-            $("#filter").on("change",
+            $("#Filter").on("change",
```
chrisjensenuk: Can you use something like this to workout the Filter Id please. Then it won't break should we rename the property.
`var filterId = "#@Html.IdFor(m => m.Filter)";`

src/Provider/Provider.Web/Orchestrators/DashboardOrchestrator.cs
```diff
@@ -31,11 +30,8 @@ public async Task<DashboardViewModel> GetDashboardViewModelAsync(long ukprn, str
 
             var filterStatus = SanitizeFilter(filter);
```
dev4valtech: The variable is not FilteringOptions, so can we rename it accordingly.

src/Provider/Provider.Web/Orchestrators/DashboardOrchestrator.cs
```diff
@@ -7,35 +7,34 @@
...
         {
             var vacancies = await GetVacanciesAsync(ukprn);
 
-            var filterStatus = SanitizeFilter(filter);
-
-            var filteredVacancies = vacancies.Where(v => 
-                    filterStatus.HasValue == false || v.Status == filterStatus.Value)
-                    .OrderByDescending(v => v.CreatedDate)
-                .ToList();
+            var filteringOptions = SanitizeFilter(filter);
```
chrisjensenuk: filteringOption singular

src/Provider/Provider.Web/Orchestrators/DashboardOrchestrator.cs
```diff
@@ -91,51 +126,28 @@ private int SanitizePage(int page, int totalVacancies)
...
-                return (VacancyStatus)status;
-
-            return null;
+            if (Enum.TryParse(typeof(FilteringOptions), filter, out var status))
+                return (FilteringOptions)status;
+            return FilteringOptions.All;
         }
 
-        private List<SelectListItem> GetFilterSelectOptions(List<VacancySummary> vacancies, VacancyStatus? filterStatus)
+        private string GetFilterHeading(int totalVacancies, FilteringOptions options)
```
chrisjensenuk: filteringOption

src/Provider/Provider.Web/Orchestrators/DashboardOrchestrator.cs
```diff
@@ -56,23 +55,59 @@ public async Task<DashboardViewModel> GetDashboardViewModelAsync(long ukprn, str
...
+                    filteredVacancies = vacancies.Where(v => v.NoOfNewApplications > 0);
+                    break;
+                case FilteringOptions.AllApplications:
+                    filteredVacancies = vacancies.Where(v =>
+                        v.NoOfSuccessfulApplications > 0 || v.NoOfUnsuccessfulApplications > 0 ||
+                        v.NoOfNewApplications > 0);
+                    break;
+                case FilteringOptions.ClosingSoon:
+                    filteredVacancies = vacancies.Where(v =>
+                        v.ClosingDate <= _timeProvider.Now.AddDays(5) && v.Status == VacancyStatus.Live);
```
chrisjensenuk: _timeProvider.Today

src/Provider/Provider.Web/Orchestrators/DashboardOrchestrator.cs
```diff
@@ -56,23 +55,59 @@ public async Task<DashboardViewModel> GetDashboardViewModelAsync(long ukprn, str
...
+                        v.NoOfSuccessfulApplications > 0 || v.NoOfUnsuccessfulApplications > 0 ||
+                        v.NoOfNewApplications > 0);
+                    break;
+                case FilteringOptions.ClosingSoon:
+                    filteredVacancies = vacancies.Where(v =>
+                        v.ClosingDate <= _timeProvider.Now.AddDays(5) && v.Status == VacancyStatus.Live);                    
+                    break;
+                case FilteringOptions.ClosingSoonWithNoApplications:
+                    filteredVacancies = vacancies.Where(v =>
+                        v.ClosingDate <= _timeProvider.Now.AddDays(5) && v.Status == VacancyStatus.Live && (v.NoOfSuccessfulApplications == 0 || v.NoOfUnsuccessfulApplications == 0 ||
```
chrisjensenuk: ditto

shomavg:
Merged extended filtering options.

PR #568, opened 2019-04-30
ER-1019 QA Anonymous employers

chrisjensenuk:
- Running profanity and banned phrase check on Anonymous employer description
- When creating the vacancy projection I am removing the EmployerWebsite and only returning the Outcode for the address.

### DIFF
--- a/src/Shared/UnitTests/Vacancies.Client/Infrastructure/Extensions/VacancyExtensionsTests.cs
+++ b/src/Shared/UnitTests/Vacancies.Client/Infrastructure/Extensions/VacancyExtensionsTests.cs
@@ -1,0 +1,159 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using AutoFixture;
+using Esfa.Recruit.Vacancies.Client.Application.Providers;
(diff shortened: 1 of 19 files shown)


src/Shared/UnitTests/Vacancies.Client/Infrastructure/Extensions/VacancyExtensionsTests.cs
```diff
@@ -0,0 +1,159 @@
...
+using Esfa.Recruit.Vacancies.Client.Infrastructure.ReferenceData.ApprenticeshipProgrammes;
+using FluentAssertions;
+using Moq;
+using Xunit;
+
+namespace Esfa.Recruit.Vacancies.Client.UnitTests.Vacancies.Client.Infrastructure.Extensions
+{
+    public class VacancyExtensionsTests
+    {
+        private const string _documentId = "doc_id";
```
sachpatel: Because this is a const could it be called `DocumentId`

chrisjensenuk: 👍

PR #569, opened 2019-04-30
Added role="menuitem" to das-navigation as requested by Analytics

sdholder:


sachpatel:
Closing as we do not need it. The `role="menuitem"` was missing on the menu recruit link on the PAS website. This is going to be corrected on PAS by the commitments team.

PR #571, opened 2019-05-01
ER-1226 - Provider user. Storing email and Ukprn

chrisjensenuk:
For Provider users we are now storing their email address and UKPRN.

Email and UKPRN get patched on every provider user log in.


shomavg:
Is it worth changing the implementation for whitelisting providers where we get the UKPRN from the routedata?

PR #588, opened 2019-05-09
ER-1150 Implements search on provider vacancies

dev4valtech:


src/Provider/Provider.Web/Orchestrators/SearchHelperOrchestrator.cs
```diff
@@ -0,0 +1,69 @@
...
+            var result = new List<string>();
+
+            if (term == null || term.Trim().Length < 3) return result;
+
+            var vacancies = await GetVacanciesAsync(ukprn);
+
+            var data = vacancies
+                .Select(v => v.Title)
+                .Where(v => v.StartsWith(term, StringComparison.OrdinalIgnoreCase))
+                .Take(maxCount);
```
shomavg: Given that we are setting to take(maxcount), do we still need to check if data.count() < maxcount?

dev4valtech: if the query returns less than max then we need to fetch more suggestions.

PR #589, opened 2019-05-10
ER-1180 Blob Storage Query Store

chrisjensenuk:
- Removed Table Storage references & classes.
- Added extra config boolean. `UseBlobStorageQueryStore`


chrisjensenuk:
This has been put on hold until after PB

PR #591, opened 2019-05-13
ER-1181_Change the survey url.

shomavg:


shomavg:
> Should these also be updated in Employer.Web?

It is for Employer.Web.

mrpmorris:
> Should these also be updated in Employer.Web?

Ignore that, I just checked and they are there

PR #594, opened 2019-05-15
ER-1197 Implements search from provider dashboard

dev4valtech:


src/Provider/Provider.Web/Views/Dashboard/Dashboard.cshtml
```diff
@@ -2,6 +2,10 @@
 
 @model DashboardViewModel
 
+@{
+    var dashboardSearchEndpointPrefixUrl = @Url.RouteUrl(RouteNames.VacanciesSearchSuggestions_Get, new { term = "" }, Context.Request.Scheme);
```
chrisjensenuk: `null` is used instead of `new { term = "" }` in Vacacnies.cshtml. Do we need to specify the term?

PR #610, opened 2019-05-28
ER-1214 employer manage notifications

dev4valtech:


src/Shared/Recruit.Vacancies.Client/Domain/Entities/UserNotificationPreferences.cs
```diff
@@ -0,0 +1,13 @@
+using System;
+
+namespace Esfa.Recruit.Vacancies.Client.Domain.Entities
+{
+    public class UserNotificationPreferences
+    {
+        // user's id
+        public Guid Id { get; set; }
+        public NotificationTypes NotificationTypes { get; set; }
```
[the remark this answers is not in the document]
dev4valtech: I have merely followed the design document. However, I figured what would the queries look like in each case:
```
db.getCollection('userNotificationPreferences').find({notificationTypes:/VacancyRejected/ })
db.getCollection('userNotificationPreferences').find({notificationTypesArr:"VacancyRejected" })
```
First one is contains query in a string field and the following one is contains query in array.

There is very subtle difference, but it would be interesting to find out when we have thousands of users.

src/Employer/Employer.Web/Filters/LevyDeclarationCheckFilter.cs
```diff
@@ -13,17 +13,18 @@ namespace Esfa.Recruit.Employer.Web.Filters
...
         private readonly ILogger<LevyDeclarationCheckFilter> _logger;
         private readonly LevyDeclarationCookieWriter _levyCookieWriter;
+        private readonly IRecruitVacancyClient _recruitVacancyClient;
 
-        public LevyDeclarationCheckFilter(IEmployerVacancyClient vacancyClient,
+        public LevyDeclarationCheckFilter(
             ILogger<LevyDeclarationCheckFilter> logger,
-            LevyDeclarationCookieWriter levyCookieWriter)
+            LevyDeclarationCookieWriter levyCookieWriter,
+            IRecruitVacancyClient recruitVacancyClient)
```
sachpatel: The LevyDeclaration check is only applicable on Employer Recruit only shirley? Why is it being put in the Recruit client. Can we get rid of all these client classes (I'm aware I am the **only** one with a problem with them but please please please)

src/Employer/Employer.Web/Orchestrators/ManageNotificationsOrchestrator.cs
```diff
@@ -0,0 +1,118 @@
...
+    public class ManageNotificationsOrchestrator : EntityValidatingOrchestrator<UserNotificationPreferences, ManageNotificationsEditModel>
+    {
+        private readonly IRecruitVacancyClient _recruitVacancyClient;
+
+        private const string NotificationTypesIsRequiredForTheFirstTime = "Choose when you’d like to receive emails";
+        private readonly EntityValidationResult _notificationTypeIsRequiredForTheFirstTime = new EntityValidationResult
+        {
+            Errors = new[] 
+            { 
+                new EntityValidationError(1100, nameof(ManageNotificationsEditModel.HasAnySubscription), NotificationTypesIsRequiredForTheFirstTime, "1100")
```
sachpatel: Can the 1100 be put in a const please.

src/Employer/Employer.Web/Views/ManageNotifications/ManageNotifications.cshtml
```diff
@@ -0,0 +1,107 @@
...
+                                    <div class="govuk-radios__item">
+                                        <input class="govuk-radios__input" value="Immediately" asp-for="NotificationFrequency" id="notification-frequency-immediately" type="radio">
+                                        <label class="govuk-label govuk-radios__label" for="notification-frequency-immediately">
+                                            Immediately
+                                        </label>
+                                    </div>
+                                    <div class="govuk-radios__item">
+                                        <input class="govuk-radios__input" value="Daily" asp-for="NotificationFrequency" id="notification-frequency-daily" type="radio" >
+                                        <label class="govuk-label govuk-radios__label" for="notification-frequency-daily">
+                                            Daily
```
sachpatel: Whoever had the idea of calling these options single one words is a genius

src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/UpdateUserNotificationPreferencesCommandHandler.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        private readonly IUserNotificationPreferencesRepository _userNotificationPreferencesRepository;
+        public UpdateUserNotificationPreferencesCommandHandler(
+            IUserNotificationPreferencesRepository userNotificationPreferencesRepository)
+        {
+            _userNotificationPreferencesRepository = userNotificationPreferencesRepository;
+        }
+        public async Task Handle(UpdateUserNotificationPreferencesCommand message, CancellationToken cancellationToken)
+        {
+            await _userNotificationPreferencesRepository.UpsertAsync(message.UserNotificationPreferences);
```
sachpatel: `return` rather than `await` here.

src/Shared/Recruit.Vacancies.Client/Application/Validation/Fluent/UserNotificationPreferencesValidator.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    public class UserNotificationPreferencesValidator : AbstractValidator<UserNotificationPreferences>
+    {
+        public UserNotificationPreferencesValidator()
+        {
+            When(n => n.NotificationTypes != NotificationTypes.None, () => 
+            {
+                RuleFor(t => t.NotificationScope)
+                    .NotNull()
+                    .WithErrorCode("1101")
```
sachpatel: So we're using this error code with two diff messages?

src/Shared/Recruit.Vacancies.Client/Domain/Entities/UserNotificationPreferences.cs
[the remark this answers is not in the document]
sachpatel: Bear in mind I think this will likely change 🤣

src/Employer/Employer.Web/Filters/LevyDeclarationCheckFilter.cs
dev4valtech: This wasn't introduced by me, seems to me like tech debt... but greater tech debt is to get rid of client classes as you mentioned and I share that sentiment.

src/Employer/Employer.Web/Views/ManageNotifications/ManageNotifications.cshtml
sachpatel: Could you use the enum value to output the display text here is what I was hinting at and the genius is me.

src/Employer/Employer.Web/Orchestrators/ManageNotificationsOrchestrator.cs
dev4valtech: It is of no value what so ever, don't think every literal needs to be a constant. It makes sense where you can give it a sensible name and make the code more readable which is not the case here.

PR #623, opened 2019-06-11
ER-1315 - Slack notifications on Live Vacancy date amendment.

sachpatel:


src/Shared/Recruit.Vacancies.Client/Application/LiveUpdateKind.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+
+namespace Esfa.Recruit.Vacancies.Client.Application
+{
+    [Flags]
+    public enum LiveUpdateKind
+    {
+        None,
+        ApplicationClosingDate,
```
chrisjensenuk: Just call this `ClosingDate`

src/Shared/Recruit.Vacancies.Client/Application/Services/INotifyVacancyUpdates.cs
```diff
@@ -6,5 +6,6 @@ namespace Esfa.Recruit.Vacancies.Client.Application.Services
     public interface INotifyVacancyUpdates
     {
         Task VacancyManuallyClosed(Vacancy vacancy);
+        Task VacancyExtended(Vacancy vacancy);
```
chrisjensenuk: If the vacancy's start date has been changed then it hasn't really been extended.  Should we just call this `LiveVacancyChanged`

src/Shared/Recruit.Vacancies.Client/Application/EventHandlers/NotifyOnVacancyActionsEventHandler.cs
```diff
@@ -36,10 +37,33 @@ public async Task Handle(VacancyClosedEvent notification, CancellationToken canc
...
             }
         }
+
+        public async Task Handle(LiveVacancyUpdatedEvent notification, CancellationToken cancellationToken)
+        {
+            try
+            {
+                var vacancy = await _vacancyRepository.GetVacancyAsync(notification.VacancyId);
+
+                switch (notification.UpdateKind)
```
chrisjensenuk: Would this be simpler as an `if`?

src/Shared/Recruit.Vacancies.Client/Application/LiveUpdateKind.cs
sachpatel: 👍

src/Shared/Recruit.Vacancies.Client/Application/Services/INotifyVacancyUpdates.cs
sachpatel: 👍

src/Shared/Recruit.Vacancies.Client/Application/EventHandlers/NotifyOnVacancyActionsEventHandler.cs
sachpatel: This is subjective.

I thought about an `if` when writing this. I opted for the `switch` with fall-through cases where applicable because it is more readable to me in a collapsed form rather than an `if` in one long line or spanning two lines or multiple `if`'s.

The idea of the switch was also to cater for the not implemented case where an update kind could be `WageAdjustment` and the combination of that with a date change.

sachpatel:
@chrisjensenuk The purpose of this PR and change is to allow us to more conveniently monitor and be aware of the use of the `Change dates on a live vacancy` feature which is actually not well designed/implemented in my opinion and I am unsure of how often this is being used. Closing a live vacancy has been seen to be used quite frequently which is interesting.

The passing of complete vacancy object into a command handler has been noted and I can understand concern but it needs an overhaul in other places as well. I understand this should be made consistent but I don't believe as part of this PR (The scope is larger than what this PR introduces). If the command handler is processing in the same process I don't see what the delay would be unless it's the possibility of retry policy caused delay further in the routine.

I am not a fan of the `LiveUpdateKind` thing I have introduced to pass the change indicator between the orchestrator through the client into a command handler but it is/was the simplest thing I could do as I was writing this with what we already have. As you have mentioned I would also much prefer to let the handler work out the changes but this is where we have a bigger issue to address as noted on the Tech Debt page.

This PR is a bit of a chicken and an egg situation as the purpose is to provide more insight. However the SlackNotification changes as well as the whole change live vacancy feature is written with the possibility of WageAdjustment being introduced. Over-engineering and writing in this way is not advised as we should keep the changes minimal and to the purpose of requirements. However this is a tech ticket and rather than further deliberation I would prefer to get this change out so we can gain insight and then make changes or remove after that.

Happy to make the other naming changes as you have suggested.

PR #627, opened 2019-06-13
ER-1294 - Employer landing page content

chrisjensenuk:
We now display quite a lot of content on the landing page when an employer has no vacancies. So I've moved this content into a partial

We now want to display the same content on DEMO and PROD so I have removed the Employer Levy feature toggle (That was short lived!)

We now display links to FAT for searching for training and providers.  I've added these url to appsettings.json

### DIFF
--- a/src/Employer/Employer.Web/Views/Dashboard/Dashboard.cshtml
+++ b/src/Employer/Employer.Web/Views/Dashboard/Dashboard.cshtml
@@ -25,7 +25,7 @@
     </div>
 </div>
 
-<partial name="_NoVacanciesContent" model="@Model" />
+<partial asp-hide="@Model.HasVacancies" name="_NoVacanciesContent" />
 
 <div asp-show="Model.HasVacancies">
     <div class="govuk-grid-row">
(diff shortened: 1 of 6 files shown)


src/Employer/Employer.Web/Views/Dashboard/Dashboard.cshtml
```diff
@@ -20,138 +19,119 @@
...
-                    You must be a levy-paying employer to post a vacancy.
-                </p>
-                <p class="govuk-body">
-                    You can clone vacancies once they've been submitted.
-                </p>
-                <p class="govuk-body">
-                    Vacancies will be approved or rejected within 24 hours.
-                </p>
-            </esfaFeatureEnabled>
+<partial name="_NoVacanciesContent" model="@Model" />
```
dev4valtech: If you use asp-show here, then no need to pass the model to the partial view. Also it makes the partial view agnostic.

chrisjensenuk: 👍 I had assumed this wouldn't work with <partial />. Turns out is does!

PR #629, opened 2019-06-13
ER-1272 content change

shomavg:


src/Shared/Recruit.Vacancies.Client/Application/Validation/Fluent/FluentVacancyValidator.cs
```diff
@@ -90,7 +90,7 @@ private void ValidateTitle()
             RuleFor(x => x.Title)
                 .Cascade(CascadeMode.StopOnFirstFailure)
                 .NotEmpty()
-                    .WithMessage("You must provide a vacancy title")
+                    .WithMessage("Enter the title of this vacancy")
```
sachpatel: I am unsure about these content changes because what it has been changed to are geared towards a web front end but this Validation layer will be re-used for the v2 vacancy api create endpoint so I think these will need to be reviewed again. Or we find a different way of setting these messages for the web from a different class within the individual products.

PR #632, opened 2019-06-14
ER 1263 vacancy summary enhancement for internal recruit api

sachpatel:
These are additional fields in the projection to support the Recruit API. There should be no breaking changes. For this to go through the environments I will be regenerating the affected views as to make it compatible for the Recruit API.

The intent is to drop the CandidateName property from the VacancyApplication subdocument type. I have made it persist-able temporarily while transitioning into the FirstName and LastName properties being introduced.

The EmployerName property was removed temporarily from the VacancySummary sub-document as it was being used for LegalEntityName but incorrectly named. Now we need a property for LegalEntityName which will be used in its place. The EmployerName property was added back in after which will map correctly from the TradingName/RegisteredName as associated with the Vacancy


src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/VacancySummariesProvider/VacancySummaryMapper.cs
```diff
@@ -13,12 +13,18 @@ internal static VacancySummary MapFromVacancySummaryAggQueryResponseDto(VacancyS
                 Id = vacSummaryDetail.VacancyGuid,
                 Title = vacSummaryDetail.Title,
                 VacancyReference = vacSummaryDetail.VacancyReference,
-                EmployerName = vacSummaryDetail.LegalEntityName,
```
dev4valtech: 💥

sachpatel: Sorted

PR #641, opened 2019-06-21
ER-919 - Provider RoATP check

chrisjensenuk:
Whenever a user accesses Provider.Web we need to check that the Provider is listed on RoATP.

This check is done in `ProviderAccountHandler` which is called on every page request. To prevent this triggering a call to the FAT API on every request I'm using the existing SetupProvider cookie to determine if this check has been done already in the session.

Assumption: All we need to do to check if a provider is in RoATP is to try and get the provider from from the FAT Api and if the provider does not exist then this means they are not in RoATP.  Is this correct?  This is how PAS does it.

### DIFF
--- a/src/Provider/Provider.Web/Middleware/ProviderAccountHandler.cs
+++ b/src/Provider/Provider.Web/Middleware/ProviderAccountHandler.cs
@@ -4,12 +4,11 @@ using System.Security.Claims;
 using System.Threading.Tasks;
 using Esfa.Recruit.Provider.Web.Configuration;
 using Esfa.Recruit.Provider.Web.Configuration.Routing;
-using Esfa.Recruit.Vacancies.Client.Application.Providers;
+using Esfa.Recruit.Vacancies.Client.Application.Configuration;
 using Esfa.Recruit.Vacancies.Client.Infrastructure.Client;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Mvc.Filters;
-using SFA.DAS.Providers.Api.Client;
 
 namespace Esfa.Recruit.Provider.Web.Middleware
 {
@@ -95,8 +94,15 @@ namespace Esfa.Recruit.Provider.Web.Middleware
             try
             {
                 var ukprnFromClaim = context.User.FindFirst(_ukprnClaimFinderPredicate).Value;
+
+                if (long.TryParse(ukprnFromClaim, out var ukprn) == false)
+                    return false;
+
+                if (ukprn == EsfaTestTrainingProvider.Ukprn)
+                    return true;
+
                 var allProviders = await _vacancyClient.GetAllTrainingProvidersAsync();
-                var provider = allProviders.SingleOrDefault(p => p.Ukprn == long.Parse(ukprnFromClaim));
+                var provider = allProviders.SingleOrDefault(p => p.Ukprn == ukprn);
                 return provider != null;
             }
             catch (Exception)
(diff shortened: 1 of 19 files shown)


sachpatel:
Looks okay. I think we might need to actually pull down the provider data and store into our referenceData. The reason is because of the reliance on the provider api might be a bit shaky based on recent events.

@mgwilliam thoughts?

Otherwise could we not do an eager pull of the providers we store in the host memory cache on app startup and the AccountHandler can just query from within that cache. I am just a little concerned about the latency added by having another web call being made as part of the auth even though you have limited that to once per session.

chrisjensenuk:
After having chatted to @mgwilliam  we are going to cache a list of providers to handle this.  I've now created `TrainingProviderSummaryProvider` (In future this may be driven from reference data).  This provider caches training providers until 6am the next day.  Its used by `Provider.Web` for authorization checking and `Employer.Web for the training provider autocomplete.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/TrainingProviderSummaryProvider/TrainingProviderSummaryProvider.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        public Task<IEnumerable<TrainingProviderSummary>> FindAllAsync()
+        {
+            return _cache.CacheAsideAsync(
+                CacheKeys.TrainingProviders,
+                _timeProvider.NextDay6am,
+                FindAllAsyncInternal);
+        }
+
+        private async Task<IEnumerable<TrainingProviderSummary>> FindAllAsyncInternal()
```
sachpatel: `FindAllInternalAsync`

src/Provider/Provider.Web/Middleware/ProviderAccountHandler.cs
```diff
@@ -81,16 +86,63 @@ private async Task<bool> HasUkprnAuthorizationAsync(AuthorizationHandlerContext
...
+            if (HasDoneOncePerAuthorizedSessionActions(context))
+                return true;
 
-            if (context.Request.Cookies[key] == null)
+            try
             {
-                await _client.SetupProviderAsync(ukprn);
-                context.Response.Cookies.Append(key, "1", EsfaCookieOptions.GetDefaultHttpCookieOption(_hostingEnvironment));
+                var ukprnFromClaim = context.User.FindFirst(_ukprnClaimFinderPredicate).Value;
+                var allProviders = await _vacancyClient.GetAllTrainingProvidersAsync();
```
sachpatel: Just a thought, haven't tried it myself but could we not throw an Authorisation exception from the ProviderAccountHandler which indicates the roatp failure instead of making potential network calls from this ErrorController.

sachpatel:
Can you enrich whats comes back from roatp in prod with 10033670 or handle passthrough f it in ProviderAccountHandler as that is the ESFA Ukprn. Otherwise we won't be able to smoke test 🚬

src/Provider/Provider.Web/Middleware/ProviderAccountHandler.cs
chrisjensenuk: Unfortunately there is no way to add context to the 403 that the `AuthorizationHandler` raises so there is no way to find out why the 403 occurred from the error controller.  We could `RedirectToRoute` directly from the handler to a known route but that would mean introducing a special case for this particular exception which would add complexity.  At the moment all 403 errors go through the error controller which I think is a good thing.  Its not ideal however I think overall its probably the best we can do.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/TrainingProviderSummaryProvider/TrainingProviderSummaryProvider.cs
chrisjensenuk: 👍  Whoops!

PR #642, opened 2019-06-21
ER-1307 improve summary block

shomavg:


src/Provider/Provider.Web/Controllers/VacancyPreviewController.cs
```diff
@@ -86,7 +88,32 @@ private void SetSectionStates(VacancyPreviewViewModel viewModel)
...
+
+        private int GetSectionStateCount(VacancyPreviewViewModel viewModel)
+        {
+            var count = 0;
+            if (CheckIfSectionIsIncomplete(viewModel.ShortDescriptionSectionState))
+                count++;
+            if (CheckIfSectionIsIncomplete(viewModel.SkillsSectionState))
+                count++;
+            if (CheckIfSectionIsIncomplete(viewModel.DescriptionsSectionState))
+                count += 3;
```
chrisjensenuk: Does this need to be incremented by 3? The descriptions all come under a single section

src/Shared/Recruit.Shared.Web/Extensions/VacancyPreviewHelper.cs
```diff
@@ -0,0 +1,28 @@
...
+namespace Esfa.Recruit.Shared.Web.Extensions
+{
+    public static class VacancyPreviewHelper
+    {
+        public static IHtmlContent ActionLink(IUrlHelper urlHelper, string linkText, string routeName, object routeValues, string screenReaderText, string linkClass, string dataAutomationId)
+        {
+            // create wrapper around text e.g. <span>linkText</span>
+            var tagBuilderInner = new TagBuilder("span");
+            tagBuilderInner.AddCssClass("govuk-visually-hidden");
+            tagBuilderInner.InnerHtml.Append(!string.IsNullOrEmpty(screenReaderText) ? HttpUtility.HtmlEncode(screenReaderText) : string.Empty);
```
chrisjensenuk: Do we need to do this check? I don't think this is ever null.

src/Shared/Recruit.Shared.Web/Extensions/VacancyPreviewHelper.cs
```diff
@@ -0,0 +1,28 @@
...
+            // create wrapper around wrapper around text e.g. <a><span>linkText</span></a>
+            var tagBuilderAtag = new TagBuilder("a");
+            tagBuilderAtag.AddCssClass(linkClass);
+            tagBuilderAtag.InnerHtml.Append(!string.IsNullOrEmpty(linkText) ? HttpUtility.HtmlEncode(linkText) : string.Empty);
```
chrisjensenuk: Do we need to do this check? I don't think this is ever null.

src/Shared/Recruit.Shared.Web/Extensions/VacancyPreviewHelper.cs
```diff
@@ -0,0 +1,28 @@
...
+
+
+namespace Esfa.Recruit.Shared.Web.Extensions
+{
+    public static class VacancyPreviewHelper
+    {
+        public static IHtmlContent ActionLink(IUrlHelper urlHelper, string linkText, string routeName, object routeValues, string screenReaderText, string linkClass, string dataAutomationId)
+        {
+            // create wrapper around text e.g. <span>linkText</span>
+            var tagBuilderInner = new TagBuilder("span");
```
chrisjensenuk: Can we call this `visuallyHiddenSpan`?

PR #643, opened 2019-06-25
ER-1288 provider user errors

dev4valtech:
I didn't realize that user signed in command is used for both provider and employer web. Also I think adding another in-process event was bit over done. So got rid of that and raising storage event straight from the command handler if usertype is Employer.


src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/UserSignedInCommandHandler.cs
```diff
@@ -50,7 +55,8 @@ private async Task UpsertUserAsync(VacancyUser user, UserType userType)
 
             await _userRepository.UpsertUserAsync(userEntity);
 
-            await _messaging.PublishEvent(new UserSignedInEvent { IdamsUserId = user.UserId } );
+            if (userType == UserType.Employer)
+                await _queueService.AddMessageAsync(new UpdateUserAccountQueueMessage { IdamsUserId = user.UserId });
```
chrisjensenuk: If this is Employer specific then should we call this something like `UpdateEmployerUserAccountQueueMessage ` Or maybe better to add the user type as a parameter to UpdateUserAccountQueueMessage so we can reuse this message should we need to update provider users in the future?

dev4valtech: Renaming as suggested as Provider Users have no association with employers.

PR #650, opened 2019-06-28
ER-889 provider agreement

chrisjensenuk:
Before we allow a vacancy to be submitted we must check the provider has a signed agreement.

Don't merge until PAS API has been updated. See this PAS PR
https://github.com/SkillsFundingAgency/das-providerapprenticeshipsservice/pull/468


chrisjensenuk:
Replaced by #655

PR #657, opened 2019-07-03
ER-1339 - Employer apprenticeship duration

chrisjensenuk:


src/Employer/Employer.Web/Orchestrators/Part1/WageOrchestrator.cs
```diff
@@ -31,17 +31,21 @@ public WageOrchestrator(IEmployerVacancyClient client, IRecruitVacancyClient vac
...
+                DurationUnit = vacancy.Wage?.DurationUnit ?? DurationUnit.Month,
                 WorkingWeekDescription = vacancy.Wage?.WorkingWeekDescription,
                 WeeklyHours = $"{vacancy.Wage?.WeeklyHours:0.##}",
                 WageType = vacancy.Wage?.WageType,
                 FixedWageYearlyAmount = vacancy.Wage?.FixedWageYearlyAmount?.AsMoney(),
                 WageAdditionalInformation = vacancy.Wage?.WageAdditionalInformation,
-                PageInfo = Utility.GetPartOnePageInfo(vacancy)
+                PageInfo = Utility.GetPartOnePageInfo(vacancy),
+                TrainingTitle = training?.Title,
+                TrainingDurationMonths = training?.Duration ?? 0
```
mrpmorris: Shouldn't this be set depending on vacancy.Wage?.DurationUnit - because it might be years?

chrisjensenuk: The duration of training courses in FAT is always months so I think we're safe here.

PR #658, opened 2019-07-03
ER-1374 📦⬆ - General sweep of safe package upgrades

sachpatel:
I have put particular regression test points relating to this upgrade in the JIRA ticket https://skillsfundingagency.atlassian.net/browse/ER-1374


sachpatel:
💥

sachpatel:
So the one package ai did a major upgrade of expecting no breaking changes that were not documented I have reverted back so all the upgrades are safe. We will have to tackle upgrading Asp.Net FluentValidation sometime in the future.

PR #661, opened 2019-07-04
Bug fix

shomavg:
Content change for ER-1272.


shomavg:
This is no more a bug.

PR #662, opened 2019-07-05
ER-1372 remap foundation and masters

mrpmorris:


src/Shared/Recruit.Shared.Web/Extensions/EnumExtensions.cs
```diff
@@ -19,8 +19,10 @@ public static string GetDisplayName(this Enum enumValue)
 
         private static readonly Dictionary<Enum, string> DisplayNames = new Dictionary<Enum, string>
         {
+#pragma warning disable CS0618 // Type or member is obsolete
             { ProgrammeLevel.FoundationDegree, "Foundation Degree" },
             { ProgrammeLevel.Masters, "Master's Degree" },
```
sachpatel: Why can't we remove this if nothing is going to be set to this? Or is this after your first round of changes?

src/Shared/Recruit.Vacancies.Client/Domain/Entities/ProgrammeLevelHelper.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System;
+
+namespace Esfa.Recruit.Vacancies.Client.Domain.Entities
+{
+    public static class ProgrammeLevelHelper
+    {
+        public static bool TryRemapFromInt(int value, out ProgrammeLevel result)
```
sachpatel: Can this method please be static as it is only being called from one place which is within this class.

src/Shared/Recruit.Vacancies.Client/Domain/Entities/ProgrammeLevelHelper.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            switch (value)
+            {
+                case 5:
```
sachpatel: I assume this is more permanent as long as the rule exists to make the modification. Reading this class and understanding the particular change from x to y is difficult. Could you please put `case 5: // PhD` or whatever 5 was supposed to be and the same for `case 7:` please.

src/Shared/UnitTests/Vacancies.Client/Domain/Entities/ProgrammeLevelHelperTests.cs
```diff
@@ -0,0 +1,92 @@
...
+            }
+        }
+
+        [Fact]
+        public void RemapFromInt_ShouldThrowArgumentException_WhenPassedAnIntWithNoCorrespondingValue()
+        {
+            int intValueToConvert = (int)ProgrammeLevel.Degree + 2;
+            Assert.Throws<ArgumentException>(() =>
+            {
+                ProgrammeLevel result = ProgrammeLevelHelper.RemapFromInt(intValueToConvert);
```
sachpatel: you don't need the `ProgrammeLevel result =` do you?

src/Shared/UnitTests/Vacancies.Client/Domain/Entities/ProgrammeLevelHelperTests.cs
```diff
@@ -0,0 +1,92 @@
...
+            });
+        }
+
+        [Fact]
+        public void TryRemapFromInt_ShouldSucceedAndReturnHigher_WhenConvertingFromFoundation5()
+        {
+            bool success = ProgrammeLevelHelper.TryRemapFromInt(5, out ProgrammeLevel result);
+            success.Should().Be(true);
```
sachpatel: There are method `BeTrue()` and `BeFalse()` if that helps.

src/Shared/Recruit.Vacancies.Client/Domain/Entities/ProgrammeLevelHelper.cs
sachpatel: Sorry meant `private static`

src/Shared/Recruit.Shared.Web/Extensions/EnumExtensions.cs
mrpmorris: I'm disabling compiler warnings for a known violation. I don't like compiler warnings, ideally we should treat warnings as errors.

sachpatel: You can put in treat warnings as errors, so long as you put it across all projects

src/Shared/Recruit.Vacancies.Client/Domain/Entities/ProgrammeLevelHelper.cs
mrpmorris: I have some tests for it, so I'll leave it public

src/Shared/Recruit.Shared.Web/Extensions/EnumExtensions.cs
mrpmorris: I know, and we haven't done that, but that's no reason to allow warnings if they can be avoided.

sachpatel: Which is the reason to dump this code, particularly in this class where it's for a UI concern.

PR #674, opened 2019-07-09
ER-1258-Remove user claims logging from provider v2

shomavg:
Remove user claims logging from provider v2


src/Provider/Provider.Web/Controllers/ErrorController.cs
```diff
@@ -148,13 +148,10 @@ private IActionResult AccessDenied()
         private void LogUserClaims()
         {
             var sb = new StringBuilder();
```
sachpatel: Nope, could you remove this whole method please we don't want to log claims as it wouldn't mean anything anymore.

Also I said in the AccessDenied action method `_logger.LogInformation($"User {User.Identity.Name} does not have service claim.");` should be changed to `_logger.LogInformation("User does not have service claim.");`

PR #675, opened 2019-07-10
ER-1368 - Header layout change in provider recruit

sachpatel:
The release for this will be arranged with the Reservations/Provider Commitments teams


src/Provider/Provider.Web/Views/Shared/_LayoutWithoutNavigation.cshtml
```diff
@@ -150,7 +136,7 @@ public string GetTitle()
         </div>
     </footer>
     <script src="/lib/govuk-frontend/dist/govuk-frontend-2.5.0.min.js"></script>
-    <script src="/lib/esfa-stacker/stacker.js"></script>
+    <script src="/lib/esfa-provider-menu/navigation.js"></script>
```
chrisjensenuk: I think we should add `asp-append-version="true"` to aid cache busting.  Add it to the link above too.

sachpatel: Have added the attribute for the navigation.js.

PR #677, opened 2019-07-11
ER-1021 changeScript for unsetting 'closedByUser' on closed closed clone vacancies

sachpatel:
I have done a check for running this script against PROD and around 67 vacancies will be treated.

The script outputs the vacancies that will and will not be affected as well as the original value that is being cleared.

This script will needs its file renaming to be ordered correctly again when it makes it into the target master and after #676 is merged.


sachpatel:
@chrisjensenuk  I think 1 minute 20 is long enough based on my run against PROD earlier. I will be monitoring whenever this goes live just to make sure if any fall just outside the net we can adjust an rerun for.

PR #682, opened 2019-07-12
ER-1414 entity changes for revoke

dev4valtech:


src/Shared/Recruit.Vacancies.Client/Domain/Entities/BlockedOrganisation.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Esfa.Recruit.Vacancies.Client.Domain.Entities
+{
+    public class BlockedOrganisation
+    {
+        public Guid Id { get; set; }
+        public OrganisationType OrganisationType { get; set; }
+        public string OrganisationId { get; set; }
+        public BlockedStatus BlockedStatus { get; set; }
+        public VacancyUser UpdatedBy { get; set; }
```
sachpatel: `UpdatedByUser`

src/Shared/Recruit.Vacancies.Client/Domain/Entities/TransferInfo.cs
```diff
@@ -0,0 +1,13 @@
+using System;
+
+namespace Esfa.Recruit.Vacancies.Client.Domain.Entities
+{
+    public class TransferInfo
+    {
+        public long Ukprn { get; set; }
+        public string ProviderName { get; set; }
```
sachpatel: I believe we should also include the `LegalEntityId` in this class unless it is the same as the one in the top level of the Vacancy document in which case `LegalEntityName` would be redundant.

dev4valtech: Legal entity can be derived from Vacancy, so I am not sure why the name was included, perhaps to make it easy to project/query, perhaps @mgwilliam can tell?

mgwilliam: @dev4valtech yes it's partly for that and also if the employer subsequently changes the legal entity - so it's basically a snapshot of the legal entity at the time of the transfer

dev4valtech: @mgwilliam does it make sense to have legal entity id as well then?

PR #684, opened 2019-07-12
ER-1413 - Repository for blocked organisations

sachpatel:


src/Shared/Recruit.Vacancies.Client/Infrastructure/Repositories/MongoDbBlockedOrganisationRepository.cs
```diff
@@ -0,0 +1,54 @@
...
+            var collection = GetCollection<BlockedOrganisation>();
+
+            var result = await RetryPolicy.ExecuteAsync(_ =>
+                collection.Find(filter).SingleOrDefaultAsync(),
+                new Context(nameof(GetAsync)));
+
+            return result;
+        }
+
+        public async Task<List<BlockedOrganisation>> GetByOrganisationIdAsync(string organisationId)
```
dev4valtech: Why not just return the most latest so a decision can be quickly made, I don't see a need for returning the list (no matter how small).

src/Shared/Recruit.Vacancies.Client/Domain/Repositories/IBlockedOrganisationQuery.cs
```diff
@@ -0,0 +1,11 @@
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Esfa.Recruit.Vacancies.Client.Domain.Entities;
+
+namespace Esfa.Recruit.Vacancies.Client.Domain.Repositories
+{
+    public interface IBlockedOrganisationQuery
+    {
+        Task<List<BlockedOrganisation>> GetByOrganisationIdAsync(string organisationId);
```
chrisjensenuk: Wouldn't we also want to pass in an organisation 'type' along with this? E.g, `Employer` or `Provider`?

src/Shared/Recruit.Vacancies.Client/Infrastructure/Repositories/MongoDbBlockedOrganisationRepository.cs
sachpatel: Yes I am writing the query to return the latest, will update

PR #685, opened 2019-07-15
ER-1361 Revoke Permission Employer.Web

chrisjensenuk:
- Now displaying the filter on the Employer Dashboard
- Added alert dismissal logic per user (Command and CommandHandler)
- Added Transferred Vacancies alert to Employer dashboard
- Added message to vacancy Manage and Edit screens

### DIFF
--- a/src/Employer/Employer.Web/Controllers/DashboardController.cs
+++ b/src/Employer/Employer.Web/Controllers/DashboardController.cs
@@ -1,9 +1,12 @@
+using System;
 using Esfa.Recruit.Employer.Web.Configuration;
 using Esfa.Recruit.Employer.Web.Configuration.Routing;
 using Esfa.Recruit.Employer.Web.Orchestrators;
 using Microsoft.AspNetCore.Mvc;
 using System.Threading.Tasks;
 using Esfa.Recruit.Employer.Web.Extensions;
+using Esfa.Recruit.Employer.Web.ViewModels.Dashboard;
+using Esfa.Recruit.Vacancies.Client.Domain.Entities;
 using Microsoft.AspNetCore.Hosting;
 
 namespace Esfa.Recruit.Employer.Web.Controllers
@@ -21,7 +24,7 @@ namespace Esfa.Recruit.Employer.Web.Controllers
         }
 
         [HttpGet("", Name = RouteNames.Dashboard_Index_Get)]
-        public async Task<IActionResult> Dashboard([FromRoute]string employerAccountId,[FromQuery] string filter, [FromQuery] int page = 1)
+        public async Task<IActionResult> Dashboard([FromRoute] string employerAccountId, [FromQuery] string filter, [FromQuery] int page = 1)
         {
             if (string.IsNullOrWhiteSpace(filter))
                 filter = Request.Cookies.GetCookie(CookieNames.DashboardFilter);
@@ -29,7 +32,7 @@ namespace Esfa.Recruit.Employer.Web.Controllers
             if (string.IsNullOrWhiteSpace(filter) == false)
                 Response.Cookies.SetSessionCookie(_hostingEnvironment, CookieNames.DashboardFilter, filter);
 
-            var vm = await _orchestrator.GetDashboardViewModelAsync(employerAccountId, filter, page);
+            var vm = await _orchestrator.GetDashboardViewModelAsync(employerAccountId, filter, page, User.ToVacancyUser());
 
             if (TempData.ContainsKey(TempDataKeys.DashboardErrorMessage))
                 vm.WarningMessage = TempData[TempDataKeys.DashboardErrorMessage].ToString();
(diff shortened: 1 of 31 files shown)


src/Employer/Employer.Web/Controllers/DashboardController.cs
```diff
@@ -39,5 +42,16 @@ public async Task<IActionResult> Dashboard([FromRoute]string employerAccountId,[
 
             return View(vm);
         }
+
+        [HttpPost("dismiss-alert", Name = RouteNames.Dashboard_DismissAlert)]
+        public async Task<IActionResult> DismissAlert([FromRoute] string employerAccountId, AlertDismissalEditModel model)
+        {
+            if (Enum.TryParse(typeof(AlertType), model.AlertType, out var alertTypeEnum))
```
dev4valtech: `Enum.TryParse<AlertType>(model.AlertType, out var result);`

src/Employer/Employer.Web/Controllers/DashboardController.cs
```diff
@@ -39,5 +42,16 @@ public async Task<IActionResult> Dashboard([FromRoute]string employerAccountId,[
...
+            {
+                await _orchestrator.DismissAlert(User.ToVacancyUser(), (AlertType)alertTypeEnum);
```
dev4valtech: If you use generic TryParse as suggested above, there is no need to cast the variable here.

src/Employer/Employer.Web/Orchestrators/DashboardOrchestrator.cs
```diff
@@ -151,5 +169,25 @@ private string GetFilterHeading(int totalVacancies, FilteringOptions filteringOp
                     return $"{totalVacancies} {filterText} {"vacancy".ToQuantity(totalVacancies, ShowQuantityAs.None)}";
             }
         }
+
+        private TransferredVacanciesAlertViewModel GetTransferredVacanciesAlertViewModel(IEnumerable<VacancySummary> vacancies, DateTime? userLastDismissedDate)
```
dev4valtech: I was thinking ahead a little bit, there are about four types of alerts that we have anticipated. Working out all the different alerts in orchestrator may make this class too big and alerts feels like cross concern to me. I was thinking inline with having a UserAlertManager with methods Get and Dismiss. Perhaps having a single partial view that can be reused when displaying alerts on top of views other then Dashboard. 

This is just an idea, I have not completely thought this through.

src/Employer/Employer.Web/Views/VacancyPreview/VacancyPreview.cshtml
```diff
@@ -36,9 +36,16 @@
 }
 <div class="govuk-grid-row">
     <div class="govuk-grid-column-full">
-        <partial name="@PartialNames.ValidationSummary" model='new ValidationSummaryViewModel { ModelState = ViewData.ModelState, OrderedFieldNames = Model.OrderedFieldNames }' />
+        <div asp-show="@Model.IsTransferred">
```
dev4valtech: So this bit of view is being repeated in ManageVacancy.cshtml. This is where a partial view with any alerts could help... again still thinking out loud on my previously mentioned idea.

src/Shared/Recruit.Vacancies.Client/Infrastructure/QueryStore/Projections/Employer/EmployerDashboard.cs
```diff
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
```
dev4valtech: 🙈

src/Employer/Employer.Web/Controllers/DashboardController.cs
chrisjensenuk: 👍

src/Employer/Employer.Web/Controllers/DashboardController.cs
chrisjensenuk: 👍

src/Employer/Employer.Web/Orchestrators/DashboardOrchestrator.cs
chrisjensenuk: I was thinking along the same lines.  I'm going to keep it as it is for now until I've implemented a few alerts and then see if its worth pulling out common behaviour.

src/Employer/Employer.Web/Views/VacancyPreview/VacancyPreview.cshtml
chrisjensenuk: Will keep in mind.

src/Shared/Recruit.Vacancies.Client/Infrastructure/QueryStore/Projections/Employer/EmployerDashboard.cs
chrisjensenuk: :+1:

chrisjensenuk:
Changes were in #697

PR #686, opened 2019-07-15
ER-1414 data update script

dev4valtech:
Since we are introducing new field closureReason, we need to backfill this field for the existing closed vacancies. 
I am using existence of ClosedByUser field to determine if the vacancy was closed manually. There is an existing issue around this, but I am hoping that will be resolved ahead of this.

### DIFF
--- a/src/Data/Recruit.ChangeScripts/010_vacancy_backfill_closureReason.js
+++ b/src/Data/Recruit.ChangeScripts/010_vacancy_backfill_closureReason.js
@@ -9,7 +9,7 @@
     }
 
     var closedVacancies = db.vacancies.aggregate([
-        { $match : {"status" : "Closed"} },
+        { $match : {"status" : "Closed"}, closureReason: {$exists: 0} },
         { $project : {
             "vacancyReference": 1, 
             "closedByUser": 1, 
--- a/src/Data/Recruit.ChangeScripts/010_vacancy_backfill_closureReason.js
+++ b/src/Data/Recruit.ChangeScripts/010_vacancy_backfill_closureReason.js
@@ -6,9 +6,7 @@
     function setClosureReason(vac, closureReason) {
         print(`updating ${vac.vacancyReference}`);
         db.vacancies.updateOne({ "_id": vac._id }, { $set: { "closureReason": closureReason} });
-    };
-
-    print("duh-1");
+    }
 
     var closedVacancies = db.vacancies.aggregate([
         { $match : {"status" : "Closed"} },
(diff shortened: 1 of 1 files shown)


mgwilliam:
@dev4valtech the existing issue that you mention... does that include cleaning up incorrectly populated "closed by" fields?

src/Data/Recruit.ChangeScripts/010_vacancy_backfill_closureReason.js
```diff
@@ -0,0 +1,48 @@
+{
+    const sleepTime = 500;
+    const autoClosure = "Auto";
+    const manualClosure = "Manual";
+
+    function setClosureReason(vac, closureReason) {
+        print(`updating ${vac.vacancyReference}`);
+        db.vacancies.updateOne({ "_id": vac._id }, { $set: { "closureReason": closureReason} });
```
sachpatel: This is potentially going to send off 100 requests in a fraction of a second. Not sure if this might cause a `request rate is large`. Have you considered using an `updateMany`

src/Data/Recruit.ChangeScripts/010_vacancy_backfill_closureReason.js
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    var closedVacancies = db.vacancies.aggregate([
+        { $match : {"status" : "Closed"} },
```
sachpatel: As a suggestion, why don't we do two separate queries, one for system closed vacancies, then we batch through them setting the closureReason to `Auto` and another query to batch through the manual.

I assume even the current one is going to take some time and we will need to do a downtime deployment. I think for that reason that half a second sleep is too little.

dev4valtech:
> @dev4valtech the existing issue that you mention... does that include cleaning up incorrectly populated "closed by" fields?

@mgwilliam yes that is correct. @sachpatel has a script to fix it, which will be deployed in a release before public beta

src/Data/Recruit.ChangeScripts/010_vacancy_backfill_closureReason.js
dev4valtech: I did few quick tests (1500 updates back to back without any halts), there weren't any errors
There are 700+ vacancies on TEST, if it is a problem it will show up on TEST. 
I had tried UpdateMany before (for some other update) and it had failed with request rates.

src/Data/Recruit.ChangeScripts/010_vacancy_backfill_closureReason.js
dev4valtech: Yes that is another way of doing it, just extra scripting will have to be done. This is simple and it works. I am expecting no more than 2500 closed vacancies by the time we come to release this. Hopefully it should take less than 10 minutes or even quicker. Anyways, don't think we need a down time.

PR #698, opened 2019-07-18
ER-1207_ER-1250_ER-1366_Flow to recruit

shomavg:


src/Employer/Employer.Web/Controllers/Part1/TitleController.cs
```diff
@@ -53,10 +69,27 @@ public async Task<IActionResult> Title(TitleEditModel m, [FromQuery] bool wizard
...
+                if (currentVacancy.TrainingProvider == null)
+                    return RedirectToRoute(RouteNames.TrainingProvider_Select_Get, new { vacancyId = response.Data });
+            }
+                
             return wizard
                 ? RedirectToRoute(RouteNames.Training_Get, new {vacancyId = response.Data})
                 : RedirectToRoute(RouteNames.Vacancy_Preview_Get);
         }
+
+        private TitleEditModel GetReferredDataFromTempData(TitleEditModel m)
```
chrisjensenuk: I don't think the models need to know about this information can you remove these properties and just work this out in the controller.  The Models should only know about that they care about.  In this case its just the information required to show/hide and set properties of a back link.

src/Employer/Employer.Web/Controllers/Part1/TitleController.cs
```diff
@@ -53,10 +69,27 @@ public async Task<IActionResult> Title(TitleEditModel m, [FromQuery] bool wizard
                 vm.PageInfo.SetWizard(wizard);
                 return View(vm);
             }
-
+            var currentVacancy = await _orchestrator.GetCurrentVacancy(m,response.Data);
```
chrisjensenuk: I don't think you need to do this here.

If `ReferredfromMA` TempData is set then all we need to do is do what the Dashboard does when it works out what route to use when its resuming a vacancy. So Once the vacancy has been created we can just find the next valid route and then redirect to there.

src/Employer/Employer.Web/Controllers/Part1/TrainingProviderController.cs
```diff
@@ -28,19 +29,27 @@ public TrainingProviderController(TrainingProviderOrchestrator orchestrator, IRe
...
+           var vm = await _orchestrator.GetSelectTrainingProviderViewModelAsync(vrm, ukprn);
+           vm.PageInfo.SetWizard(wizard);
+
+           if (string.IsNullOrWhiteSpace(clear) == false)
+           {
+               vm.Ukprn = string.Empty;
+               vm.TrainingProviderSearch = string.Empty;
+               vm.IsTrainingProviderSelected = true;
+           }
+           vm.ReferredFromMAHome_FromSavedFavourites = ShowReferredFromMABackLink();
```
chrisjensenuk: simplify this terms of showing/hiding a back link.  The models and views shouldn't care about MA and favourites etc.

src/Employer/Employer.Web/Orchestrators/Part1/TitleOrchestrator.cs
```diff
@@ -84,6 +85,14 @@ public async Task<TitleViewModel> GetTitleViewModelAsync(TitleEditModel m)
 
         public async Task<OrchestratorResponse<Guid>> PostTitleEditModelAsync(TitleEditModel m, VacancyUser user)
         {
+            TrainingProvider provider = null;
+            IApprenticeshipProgramme programme = null;
+            if (IsReferredFromMaHomeFavourites(m))
+            {
+                provider = await GetProvider(m.ReferredFromMAHome_UKPRN);
+                programme = await GetProgramme(m.ReferredFromMAHome_ProgrammeId);
```
chrisjensenuk: What happen's if ether of these returns null?

src/Employer/Employer.Web/ViewModels/Part1/Title/TitleEditModel.cs
```diff
@@ -8,10 +8,12 @@ public class TitleEditModel
...
         public string EmployerAccountId { get; set; }
-
         [FromRoute]
         public Guid? VacancyId { get; set; }
-
         public string Title { get; set; }
+        public string ReferredFromMAHome_UKPRN { get; set; }
+        public string ReferredFromMAHome_ProgrammeId { get; set; }
+        public bool ReferredFromMAHome { get; set; }
+        public bool ReferredFromMAHome_FromSavedFavourites;
```
chrisjensenuk: This model shouldn't care about this stuff. We've got it in TempData.  Maybe add extra parameters to the Orchestrator function to pass in programmId and ukprn as optional parameters if we have them in TempData.

src/Employer/Employer.Web/ViewModels/Part1/NumberOfPositions/NumberOfPositionsViewModel.cs
```diff
@@ -10,9 +10,12 @@ public class NumberOfPositionsViewModel : VacancyRouteModel
         public PartOnePageInfoViewModel PageInfo { get; set; }
         public ReviewSummaryViewModel Review { get; set; } = new ReviewSummaryViewModel();
         public string NumberOfPositions { get; set; }
+        public bool FromMaSavedFavourites { get; set; }
```
chrisjensenuk: The model shouldn't care about MA. Just focus on the minimal amount of stuff you need to render the UI elements.

src/Employer/Employer.Web/Orchestrators/Part1/TitleOrchestrator.cs
shomavg: Just null would get passed in.

src/Employer/Employer.Web/Views/Dashboard/Dashboard.cshtml
```diff
@@ -84,44 +84,43 @@
...
         </div>
     </div>
 </div>
+</div>
 
 @section FooterJS
 {
-<script nws-csp-add-nonce="true">
-    var filterId = "#@Html.IdFor(m => m.Filter)";
+<script nws-csp-add-nonce="true">var filterId = "#@Html.IdFor(m => m.Filter)";
```
chrisjensenuk: As discussed. Can you undo these formatting changes.  I think the only change we should have is the change to the back link.

src/Employer/Employer.Web/Controllers/EmployerVacancyController.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            _employerVacancyOrchestrator = employerVacancyOrchestrator;
+        }
+
+        [HttpGet("employer-create-vacancy", Name = RouteNames.EmployerCreateVacancy_Get)]        
+        public async Task<IActionResult> CreateVacancy([FromQuery]string ukprn, [FromQuery]string programmeId)
+        {
+            TempData[TempDataKeys.ReferredFromMa] = true;
+
+            if(string.IsNullOrWhiteSpace(ukprn) && string.IsNullOrWhiteSpace(programmeId))
```
chrisjensenuk: Can you get rid of this check and just get _employerVacancyOrchestrator.GetProviderUkprn(ukprn) and _employerVacancyOrchestrator.GetProgrammeId(programmeId); to return immediately if values are null. That means we only have the simple IF statement at the bottom to work out the returned route.

Also by doing this it will remove the bug you have by coming from the MA Home page after you have come from the Favourite homepage as it doesn't clear the cookies.

src/Employer/Employer.Web/Controllers/Part1/TitleController.cs
```diff
@@ -21,27 +24,55 @@ public TitleController(TitleOrchestrator orchestrator)
...
         
         [HttpGet("create-vacancy", Name = RouteNames.CreateVacancy_Get)]
-        public IActionResult Title()
+        public async Task<IActionResult> Title([FromRoute] string employerAccountId)
         {
-            var vm = _orchestrator.GetTitleViewModel();
+            var vm = await _orchestrator.GetTitleViewModel(employerAccountId);
             vm.PageInfo.SetWizard();
+            GetReferredDataFromTempData(vm);
+            vm.ReturnToMALinkText = SetBackText(vm);
```
chrisjensenuk: This is odd. vm.ReturnToMALinkText = SetBackText({required variables})

Also, as discussed, the view doesn't care about MA. rename to `BackLinkText`

src/Employer/Employer.Web/Controllers/Part1/TitleController.cs
```diff
@@ -21,27 +24,55 @@ public TitleController(TitleOrchestrator orchestrator)
...
+            vm.ReturnToMALink = SetBackLinkRoute(vm);
```
chrisjensenuk: same comments from above comment apply.

src/Employer/Employer.Web/Controllers/Part1/TitleController.cs
```diff
@@ -54,9 +84,40 @@ public async Task<IActionResult> Title(TitleEditModel m, [FromQuery] bool wizard
                 return View(vm);
             }
 
+            if (m.ReferredFromSavedFavourites)
+            {
+                TempData[TempDataKeys.ReferredUkprn + response.Data] = TempData[TempDataKeys.ReferredUkprn];
```
chrisjensenuk: I think the creation of `TempDataKeys.ReferredUkprn + response.Data` would probably be better as a function off of TempDataKeys then you can share the logic between the controllers. E.g. `TempDataKeys.GetReferredUkprnKeyForVacancy(vacancyId)`

Obviously do the same for referred programmeId

src/Employer/Employer.Web/Controllers/Part1/TrainingProviderController.cs
```diff
@@ -20,25 +20,33 @@ public class TrainingProviderController : Controller
...
 
-            if (string.IsNullOrWhiteSpace(clear) == false)
-            {
-                vm.Ukprn = string.Empty;
-                vm.TrainingProviderSearch = string.Empty;
-                vm.IsTrainingProviderSelected = true;
-            }
-            
-            return View(vm);
+        private bool ShowReferredFromMaBackLink(Guid vacancyId)
```
chrisjensenuk: Return a string instead of bool and return the route name then you won't need to double up of IF statements.  Rename to `GetBackLinkRoute()`.

src/Employer/Employer.Web/Orchestrators/EmployerVacancyOrchestrator.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        public EmployerVacancyOrchestrator(IRecruitVacancyClient vacancyClient)
+        {
+            _vacancyClient = vacancyClient;
+        }
+        public async Task<TrainingProviderSummary> GetProviderUkprn(string ukprn)
+        {
+            if (string.IsNullOrWhiteSpace(ukprn))
+                return null;
+            long.TryParse(ukprn, out long validUkprn);
```
chrisjensenuk:
```
if(long.TryParse(ukprn, out long validUkprn) == false)
  return null;
```
Otherwise you are still doing the search.

src/Employer/Employer.Web/Orchestrators/Part1/TitleOrchestrator.cs
```diff
@@ -119,5 +129,19 @@ public async Task<OrchestratorResponse<Guid>> PostTitleEditModelAsync(TitleEditM
             mappings.Add(e => e.Title, vm => vm.Title);
             return mappings;
         }
+
+        private async Task<TrainingProvider> GetProvider(string ukprn)
+        {
+            long.TryParse(ukprn, out long validUkprn);
```
chrisjensenuk:
```
if(long.TryParse(ukprn, out long validUkprn) == false)
   return null;
```
Then you won't need the IF below

src/Employer/Employer.Web/ViewModels/Part1/Title/TitleViewModel.cs
```diff
@@ -14,7 +21,7 @@ public class TitleViewModel : TitleEditModel
         public string FormPostRouteName => VacancyId.HasValue ? RouteNames.Title_Post : RouteNames.CreateVacancy_Post;
         public PartOnePageInfoViewModel PageInfo { get; set; }
         public bool HasCloneableVacancies { get; set; }
-        public string BackLink =>
-            HasCloneableVacancies ? RouteNames.Dashboard_Index_Get : RouteNames.CreateVacancyOptions_Get;
+        public string BackLinkText { get; set; }
+        public string BackLinkRoute { get; set; }
```
chrisjensenuk: I don't think you are taking into account the hasCloneableVacancies logic that you deleted from the original BackLink in the view model. This also needs to be taken into account the in SetBackLinkRoute() function.

src/Employer/Employer.Web/Controllers/Part1/TitleController.cs
```diff
@@ -54,9 +84,40 @@ public async Task<IActionResult> Title(TitleEditModel m, [FromQuery] bool wizard
                 return View(vm);
             }
 
+            if (m.ReferredFromSavedFavourites)
```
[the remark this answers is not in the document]
shomavg: This bit of code only gets executed when a vacancy is created...

src/Employer/Employer.Web/Configuration/TempDataKeys.cs
```diff
@@ -7,5 +7,9 @@ public static class TempDataKeys
         public const string Skills = "Skills";
         public const string ApplicationReviewStatusInfoMessage = "ApplicationReviewStatus_InfoMessage";
         public const string VacancyClosedMessage = "VacancyClosedMessage";
+        public const string ReferredFromMAHome = "ReferredFromMAHome";
+        public const string ReferredFromMAHome_UKPRN = "ReferredFromMAHome_UKPRN";
+        public const string ReferredFromMAHome_ProgrammeId = "ReferredFromMAHome_ProgrammeId";
+        public const string ReferredFromMAHome_FromSavedFavourites = "ReferredFromMAHome_FromSavedFavourites";
```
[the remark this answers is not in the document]
shomavg: 👍

src/Employer/Employer.Web/ViewModels/Part1/Title/TitleViewModel.cs
```diff
@@ -16,5 +17,8 @@ public class TitleViewModel : TitleEditModel
         public bool HasCloneableVacancies { get; set; }
         public string BackLink =>
             HasCloneableVacancies ? RouteNames.Dashboard_Index_Get : RouteNames.CreateVacancyOptions_Get;
+
+        public bool ShowReturnToMALink { get; set; }
```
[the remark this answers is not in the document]
shomavg: the titleViewModel is no more inheriting from editmodel... this will solve this.

src/Employer/Employer.Web/Views/Part1/Title/Title.cshtml
```diff
@@ -1,12 +1,23 @@
...
-        <a asp-show="@Model.PageInfo.IsWizard" asp-route="@Model.BackLink" asp-route-wizard="@Model.PageInfo.IsWizard" esfa-automation="link-back" class="govuk-back-link govuk-!-margin-top-0 govuk-!-margin-bottom-6">Back</a>
         <partial name="@PartialNames.ValidationSummary" model='new ValidationSummaryViewModel { ModelState = ViewData.ModelState, OrderedFieldNames = Model.OrderedFieldNames}' />
         <partial name="@PartialNames.ReviewSummary" for="Review" />
-        <h1 class="govuk-heading-xl govuk-!-margin-bottom-6">What do you want to call this vacancy?</h1>
+        <esfaFeatureDisabled name="@FeatureNames.AllowLevyPayingEmployersOnly">
+            <a asp-show="@Model.ShowReturnToMALink" href="@Model.ReturnToMALink" class="govuk-back-link govuk-!-margin-top-0 govuk-!-margin-bottom-6">@Model.ReturnToMALinkText</a>
+            <a asp-hide="@Model.ShowReturnToMALink" asp-route="@RouteNames.Dashboard_Index_Get" class="govuk-back-link govuk-!-margin-top-0 govuk-!-margin-bottom-6">Return to your vacancies</a>
+        </esfaFeatureDisabled>
+        <h1 class="govuk-heading-xl govuk-!-margin-bottom-6">Vacancy summary</h1>
```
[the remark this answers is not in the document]
shomavg: this has been fixed.

src/Employer/Employer.Web/ViewModels/Part1/Title/TitleEditModel.cs
shomavg: As discussed, this will stay in the editmodel instead of adding more parameters to the function.

shomavg:
> I think there is some logic you have deleted and not taken into account elsewhere and a couple of other things.

the logic has been changed to 'Return to your vacancies'-once a vacancy has been created. will get this confirmed from Jithesh

shomavg:
> Much cleaner and simpler 👍 Just added a comment on possible improvement.
> 
> I also found a bug.
> 
> 1. From Favourites create a vacancy
> 2. Provide a title and click next. Progress to TrainingProvider or Position
> 3. Cancel and go back to recruit dashboard.
> 4. Click 'Createa Vacancy'
> 5. The Title back button says "Back to your saved favourites" which is wrong.

I have fixed this bug by explicitly removing the values from the tempdata.

src/Employer/Employer.Web/Orchestrators/Part1/TitleOrchestrator.cs
```diff
@@ -94,7 +101,10 @@ public async Task<OrchestratorResponse<Guid>> PostTitleEditModelAsync(TitleEditM
                 return await ValidateAndExecute(
                     newVacancy, 
                     v => _vacancyClient.Validate(v, ValidationRules),
-                    async v => await _client.CreateVacancyAsync(m.Title, m.EmployerAccountId, user));
+                    async v =>
+                    {
```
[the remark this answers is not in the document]
chrisjensenuk: I meant inside the
```
async v =>
{
   [here]
   return await _client.CreateVacancyAsync(m.Title, m.EmployerAccountId, user, provider, programme?.Id);
}
```

PR #701, opened 2019-07-19
ER-1351 SLA reports release branch

mrpmorris:


src/Shared/Recruit.Vacancies.Client/Infrastructure/Reports/QaApplicationsReportStrategy.cs
```diff
@@ -0,0 +1,224 @@
...
+            IApprenticeshipProgrammeProvider programmeProvider,
+            ITimeProvider timeProvider,
+            ILogger<QaApplicationsReportStrategy> logger)
+            : base(loggerFactory, MongoDbNames.RecruitDb, MongoDbCollectionNames.VacancyReviews, details)
+        {
+            _programmeProvider = programmeProvider;
+            _timeProvider = timeProvider;
+            _logger = logger;
+            _collection = GetCollection<BsonDocument>();
+            RetryPolicy = MongoDbRetryPolicy.GetConnectionRetryPolicy(_logger);
```
chrisjensenuk: Not sure this is right. What happens if a `MongoException` is thrown? (See other comment in Mongo DB Retry comment)

src/Shared/Recruit.Vacancies.Client/Infrastructure/Mongo/MongoDbRetryPolicy.cs
```diff
@@ -10,14 +10,25 @@ public static class MongoDbRetryPolicy
     {
         public static RetryPolicy GetRetryPolicy(ILogger logger)
         {
-            return Policy
-                .Handle<MongoException>()
+            return AddWaitAndRetry(Policy.Handle<MongoException>(), logger);
+        }
+
+        public static RetryPolicy GetConnectionRetryPolicy(ILogger logger)
```
chrisjensenuk: I'm not sure having two different policies is the way to go here.  I think you can just add this to the exisitng policy
```
.Handle<MongoException>()
.Or<MongoConnectionClosedException>()
.....
```

src/Shared/Recruit.Vacancies.Client/Infrastructure/Reports/CsvBuilder.cs
```diff
@@ -43,49 +48,42 @@ private static CsvWriter GetCsvWriter(TextWriter streamWriter)
...
-                csv.WriteField("");
-            }
-            csv.NextRecord();
-
-            WriteEmptyRow(csv, fieldCount);
-            WriteEmptyRow(csv, fieldCount);
-
-            csv.WriteField("Date");
-            csv.WriteField("Total_Number_Of_Applications");
+            WriteEmptyRow(csv, 0);
```
chrisjensenuk: Guess you didn't see my Slack reply.
```
The reason I didn't call WriteEmptyRow is that it isn't an empty row.  Its starts from i = 1 because the first field says "PROTECT"
```

src/Shared/Recruit.Vacancies.Client/Infrastructure/Reports/CsvBuilder.cs
```diff
@@ -43,49 +48,42 @@ private static CsvWriter GetCsvWriter(TextWriter streamWriter)
...
-            WriteTotalHeader(csv, reportDate, 0, 2);
+            WriteTotalHeader(csv, headers);
         }
 
-        private void WriteTotalHeader(CsvWriter csv, DateTime reportDate, int total, int fieldCount)
+        private void WriteTotalHeader(
+            CsvWriter csv,
+            IEnumerable<KeyValuePair<string, string>> headers)
         {
+            headers = headers ?? new KeyValuePair<string, string>[0];
```
chrisjensenuk: According to Mark's Slack message.  This needs to say `OFFICIAL` now.  Can you double-check and change.  Note that this will also update the reports for Provider.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Reports/ProviderApplicationsReportStrategy.cs
```diff
@@ -122,9 +122,14 @@ private async Task<string> GetProviderApplicationsAsync(long ukprn, DateTime fro
...
             var dotNetFriendlyResults = results.Select(BsonTypeMapper.MapToDotNetValue);
-            var result = JsonConvert.SerializeObject(dotNetFriendlyResults);
-            
-            return result;
+            var data = JsonConvert.SerializeObject(dotNetFriendlyResults);
+
+            var headers = new List<KeyValuePair<string, string>>
+            {
+                new KeyValuePair<string, string>("Date", _timeProvider.Now.ToLocalTime().ToString("dd/MM/yyyy hh:mm:ss")),
+                new KeyValuePair<string, string>("Total_Number_Of_Notifications", results.Count.ToString())
```
chrisjensenuk: Typo `Total_Number_Of_Applications`

src/Shared/Recruit.Vacancies.Client/Infrastructure/Reports/CsvBuilder.cs
```diff
@@ -43,49 +48,42 @@ private static CsvWriter GetCsvWriter(TextWriter streamWriter)
...
-            for (var i = 2; i < fieldCount; i++)
+            if (headers.Any())
             {
-                csv.WriteField("");
-            }
-            csv.NextRecord();
-
-            csv.WriteField(reportDate.ToString(DateTimeFormat));
-            csv.WriteField(total);
+                foreach (var kvp in headers)
```
chrisjensenuk: I think you need to pad out the rows with empty fields so each row has the same number of columns

src/Shared/Recruit.Vacancies.Client/Infrastructure/Reports/ReportService.cs
```diff
@@ -58,15 +58,16 @@ public async Task GenerateReportAsync(Guid reportId)
...
+                var reportStrategyResult = await reportStrategy.GetReportDataAsync(report.Parameters);
 
                 report = await _reportRepository.GetReportAsync(reportId);
                 if (report.Status == ReportStatus.Generated)
                 {
                     _logger.LogInformation("Report: {reportId} already generated in the meantime. Ignoring.", reportId);
                     return;
                 }
-                report.Data = data;
+                report.Headers = reportStrategyResult.Headers;
```
chrisjensenuk: What about existing Provider reports won't Headers be null? I'm not sure about this `Report.Header` field.  Can't we just pass in a per-report strategy/delegate for headers when we generate the report data?

src/Shared/Recruit.Vacancies.Client/Infrastructure/Reports/ReportService.cs
```diff
@@ -93,7 +94,7 @@ public void WriteReportAsCsv(Stream stream, Report report)
 
             var results = JArray.Parse(report.Data);
 
-            _csvBuilder.WriteCsvToStream(stream, results, report.RequestedOn.ToUkTime(), reportStrategy.ResolveFormat);
+            _csvBuilder.WriteCsvToStream(stream, results, report.RequestedOn.ToLocalTime(), reportStrategy.ResolveFormat);
```
chrisjensenuk: Can you change this back please.  ToLocalTime is no longer used and ToUkTime should now be used.
The App Services run in UTC so Local Time is meaningless.
See this PR for info.  https://github.com/SkillsFundingAgency/das-recruit/pull/621

src/Shared/Recruit.Vacancies.Client/Infrastructure/Reports/QaApplicationsReportStrategy.cs
```diff
@@ -0,0 +1,224 @@
...
+
+            var dotNetFriendlyResults = results
+                .OrderBy(x => x[Column.DateSubmitted].ToUniversalTime())
+                .ThenBy(x => x[Column.VacancyReferenceNumber].ToInt64())
+                .Select(BsonTypeMapper.MapToDotNetValue);
+
+            var data = JsonConvert.SerializeObject(dotNetFriendlyResults);
+            var headers = new List<KeyValuePair<string, string>>
+            {
+                new KeyValuePair<string, string>("Date", _timeProvider.Now.ToLocalTime().ToString("dd/MM/yyyy hh:mm:ss"))
```
chrisjensenuk: ToUkTime

src/Shared/Recruit.Vacancies.Client/Infrastructure/Reports/ProviderApplicationsReportStrategy.cs
```diff
@@ -122,9 +122,14 @@ private async Task<string> GetProviderApplicationsAsync(long ukprn, DateTime fro
...
 
             var dotNetFriendlyResults = results.Select(BsonTypeMapper.MapToDotNetValue);
-            var result = JsonConvert.SerializeObject(dotNetFriendlyResults);
-            
-            return result;
+            var data = JsonConvert.SerializeObject(dotNetFriendlyResults);
+
+            var headers = new List<KeyValuePair<string, string>>
+            {
+                new KeyValuePair<string, string>("Date", _timeProvider.Now.ToLocalTime().ToString("dd/MM/yyyy hh:mm:ss")),
```
chrisjensenuk: ToUkTime should be used.

src/QA/QA.Web/Orchestrators/Reports/ReportDashboardOrchestrator.cs
```diff
@@ -0,0 +1,63 @@
...
+            {
+                ProcessingCount = reports.Count(r => r.IsProcessing),
+                Reports = reports
+                    .OrderByDescending(r => r.RequestedOn)
+                    .Select(r => new ReportRowViewModel
+                    {
+                        ReportId = r.Id,
+                        ReportName = r.ReportName,
+                        DownloadCount = r.DownloadCount,
+                        CreatedDate = r.RequestedOn.ToLocalTime().AsGdsDateTime(),
```
chrisjensenuk: ToUkTime should be used

src/QA/QA.Web/Orchestrators/Reports/ApplicationsReportOrchestrator.cs
```diff
@@ -0,0 +1,74 @@
...
+                    break;
+                case DateRangeType.Custom:
+                    fromDate = model.FromDate.AsDateTimeUk(DateTimeStyles.AssumeLocal).Value.Date;
+                    toDateInclusive = model.ToDate.AsDateTimeUk(DateTimeStyles.AssumeLocal).Value.Date.AddDays(1).AddTicks(-1);
+                    break;
+                default:
+                    throw new Exception($"Cannot handle this date range type:{model.DateRange.ToString()}");
+            }
+
+            var reportName = $"{fromDate.ToLocalTime().AsGdsDate()} to {toDateInclusive.ToLocalTime().AsGdsDate()}";
```
chrisjensenuk: ToUkTime should be used

src/Shared/Recruit.Vacancies.Client/Infrastructure/Reports/ReportStrategyResult.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.Collections.Generic;
+
+namespace Esfa.Recruit.Vacancies.Client.Infrastructure.Reports
+{
+    public class ReportStrategyResult
```
chrisjensenuk: If we pass in a strategy for creating the header-per-report type then we won't need this added complication

src/QA/QA.Web/Views/Dashboard/Index.cshtml
```diff
@@ -128,4 +128,14 @@
...
     </div>
-</div>
\ No newline at end of file
+</div>
+
+<hr class="govuk-section-break govuk-section-break--xl govuk-section-break--visible" />
+<div class="govuk-grid-row">
+    <div class="govuk-grid-column-one-third">
+        <h3 class="govuk-heading-s">
+            <a class="govuk-link  govuk-link--no-visited-state" asp-route="@RouteNames.ReportDashboard_Get">Reports</a>
```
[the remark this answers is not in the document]
mrpmorris: 3 spaces everywhere else

src/QA/QA.Web/QA.Web.csproj
```diff
@@ -8,22 +8,22 @@
...
+    <PackageReference Include="NWebsec.AspNetCore.Middleware" Version="2.0.0" />
     <PackageReference Include="NWebsec.AspNetCore.Mvc.TagHelpers" Version="2.0.0" />
-    <PackageReference Include="Humanizer.Core.uk" Version="2.6.2"/>
+    <PackageReference Include="Humanizer.Core.uk" Version="2.6.2" />
   </ItemGroup>
   <ItemGroup>
-    <Content Update="nlog.config" CopyToOutputDirectory="PreserveNewest"/>
-    <None Update="localhost.pfx" CopyToOutputDirectory="PreserveNewest"/>
+    <Content Update="nlog.config" CopyToOutputDirectory="PreserveNewest" />
+    <None Update="localhost.pfx" CopyToOutputDirectory="PreserveNewest" />
```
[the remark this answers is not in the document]
mrpmorris: Was already there, VS has just added a space before the /

src/Shared/Recruit.Vacancies.Client/Infrastructure/Mongo/MongoDbRetryPolicy.cs
mrpmorris: MongoConncectionClosedException is descended from MongoException. 
What I want to do here is only try again for the transient error experienced when the server cannot be reached. The code itself handle the Timeout exception (no more RUs), and any other error is considered non-recoverable and aborts.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Reports/CsvBuilder.cs
mrpmorris: WriteEmptyRow(csv, 0) will write zero commas and then move to the next line. You don't need a comma per empty cell on these rows.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Reports/ReportStrategyResult.cs
mrpmorris: The report strategy generates all of the data (both in the result set and in the header) but just identifies that some of the data is for the header section and some is for the data section.

This data is stored in the Report in the DB so is only ever calculated once.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Reports/CsvBuilder.cs
chrisjensenuk: The brief for the Provider report was to make it as close to the existing v1 report as possible. This is changing the format.  We can keep this change (I agree having extra commas is not required) and fix if anyone complains. :+1:

In that case I think you'll need to remove one the the `WriteEmptyRow(csv, 0);`  calls I think its only supposed to be 2 empty lines.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Reports/ReportStrategyResult.cs
chrisjensenuk: I was just wondering if we should have a new "render strategy" per report type which would calculate and add headers and the like at time of download. I'm wondering if the concept of a 'header' is special enough to warrant its own persisted property.  Its just speculation at the moment and as we only have 2 reports both of which require a header then this solution is fine.  :+1:

src/Shared/Recruit.Vacancies.Client/Infrastructure/Reports/CsvBuilder.cs
mrpmorris:
```
            csv.WriteField("PROTECT");

            WriteEmptyRow(csv, 0);
            WriteEmptyRow(csv, 0);
            WriteEmptyRow(csv, 0);
```

That is 1 line ending for PROTEXT, then 2 empty lines :)

src/Shared/Recruit.Vacancies.Client/Infrastructure/Mongo/MongoDbRetryPolicy.cs
chrisjensenuk: 👍

PR #702, opened 2019-07-19
ER-1411 provider access revoked error page

sachpatel:
I haven't used the client classes as I would like to start moving away from adding new functionality into them.

### DIFF
--- a/src/Provider/Provider.Web/Controllers/ErrorController.cs
+++ b/src/Provider/Provider.Web/Controllers/ErrorController.cs
@@ -95,7 +95,7 @@ namespace Esfa.Recruit.Provider.Web.Controllers
                 }
 
                 if (exception is AuthorisationException)
-                {   
+                {
                     return AccessDenied();
                 }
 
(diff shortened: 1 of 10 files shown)


src/Provider/Provider.Web/Controllers/ErrorController.cs
```diff
@@ -135,6 +135,11 @@ private IActionResult MissingPermissions(long ukprn)
 
         private IActionResult AccessDenied()
         {
+            if (TempData.ContainsKey(TempDataKeys.IsBlockedProvider) && (bool)TempData.Peek(TempDataKeys.IsBlockedProvider))
```
chrisjensenuk: I'm not fussed which way we work out why the access was denied.  Either by rerunning the authorisation logic in the Error controller to work out why it failed (which is what we currently did) or by passing in temp data which you have done here.  Now that the 'authorisation' has got that bit more complex its probably best to pass in temp data for the other reasons too. Can you/have you updated the tech debt page so we can update the AccessDenied function so that it only uses TempData for the other access denied reasons? Such as "missing servie claim," "missing idams claim", "not listed in roatp", "general catch all access denied".

PR #705, opened 2019-07-23
ER-1381 Grade must be 1-9 when Qualification is a GCSE

mrpmorris:


src/Shared/Recruit.Vacancies.Client/Application/Validation/Fluent/QualificationValidator.cs
```diff
@@ -44,6 +44,16 @@ public QualificationValidator(long ruleId, IList<string> qualifications)
                     .WithErrorCode("6")
                 .WithRuleId(ruleId);
 
+            When(x => x.QualificationType != null && x.QualificationType.Contains("GCSE"), () =>
```
sachpatel: There is already a check for qualification type not being empty at the top of the file. Shouldn't need the !- null check

PR #716, opened 2019-07-29
Added ip filter range variable handling

mwevill:


sachpatel:
Hi @mwevill @jack-education  , When will this have to be integrated. Our next release might not be done using the master branch so not sure this PR should be merged into master. Let me know.

mwevill:
@sachpatel This change allows us to keep the ip restriction rules for the cosmosdb applied during releases. They are currently removed when a new release is pushed this is a security risk. While it is a none breaking change to be effective a new parameter needs adding to the release in azure devops which means any subsequent releases will need this change. Please advise which branch you would like this merged to and I will re create the PR,

dev4valtech:
> @sachpatel This change allows us to keep the ip restriction rules for the cosmosdb applied during releases. They are currently removed when a new release is pushed this is a security risk. While it is a none breaking change to be effective a new parameter needs adding to the release in azure devops which means any subsequent releases will need this change. Please advise which branch you would like this merged to and I will re create the PR,

@mwevill can you please add the required parameters to the release definition and merge this into master.

PR #725, opened 2019-08-01
ER-1414 - Fix closureReason being set to undefined

chrisjensenuk:


src/Data/Recruit.ChangeScripts/010_vacancy_backfill_closureReason.js
```diff
@@ -3,13 +3,17 @@
...
+    function setClosureReason(vac) {
         print(`updating ${vac.vacancyReference}`);
-        db.vacancies.updateOne({ "_id": vac._id }, { $set: { "closureReason": closureReason} });
+        db.vacancies.updateOne({ "_id": vac._id }, { $set: { "closureReason": vac.closureReason} });
     }
 
     var closedVacancies = db.vacancies.aggregate([
-        { $match : {"status" : "Closed", "closureReason": {$exists: 0} }},
+        { $match : {"status" : "Closed", 
+            $or : [
```
dev4valtech: I think it will be simpler to avoid any check on closureReason field, as we need to update all the closed vacancies.

chrisjensenuk: I think we still need this check as we'd be overwriting any future withdrawn vacancies every time we deploy.

PR #729, opened 2019-08-05
ER-1436 get distinct vacancy reference from application reviews refactor to use agg pipeline

sachpatel:


src/Jobs/Recruit.Vacancies.Jobs/Triggers/QueueTriggers/SpikeQueueTrigger.cs
```diff
@@ -33,7 +34,8 @@ public async Task SpikeAsync([QueueTrigger("test-queue", Connection = "QueueStor
             }
 
             var employerAccountId = Environment.GetEnvironmentVariable("EmployerAccount");
-            var result = await _query.GetEmployerOwnedVacancySummariesByEmployerAccountAsync(employerAccountId);
+            //var result = await _query.GetEmployerOwnedVacancySummariesByEmployerAccountAsync(employerAccountId);
```
mrpmorris: No need to comment out old code, we have source control

PR #752, opened 2019-08-19
ER-1387 Implements notification for withdrawn vacancies

dev4valtech:


src/Shared/Recruit.Vacancies.Client/Application/EventHandlers/NotifyOnVacancyActionsEventHandler.cs
```diff
@@ -5,41 +5,54 @@
...
             _logger = logger;
+            _communicationQueueService = communicationQueueService;
         }
 
         public async Task Handle(VacancyClosedEvent notification, CancellationToken cancellationToken)
         {
-            try
+            var vacancy = await _vacancyRepository.GetVacancyAsync(notification.VacancyId);
+
+            if (vacancy.ClosureReason == ClosureReason.WithdrawnByQa)
```
chrisjensenuk: Should this go inside a try/catch like the Manual closure below?

src/Shared/Recruit.Vacancies.Client/Application/EventHandlers/NotifyOnVacancyActionsEventHandler.cs
```diff
@@ -5,41 +5,54 @@
...
 using System.Threading.Tasks;
+using Communication.Types;
+using Esfa.Recruit.Vacancies.Client.Application.Communications;
 using Esfa.Recruit.Vacancies.Client.Domain.Repositories;
 using Esfa.Recruit.Vacancies.Client.Domain.Entities;
+using Esfa.Recruit.Vacancies.Client.Infrastructure.StorageQueue;
 
 namespace Esfa.Recruit.Vacancies.Client.Application.EventHandlers
 {
     public class NotifyOnVacancyActionsEventHandler : INotificationHandler<VacancyClosedEvent>,
```
sachpatel: The handling stuff you have addressed in this class should go in the VacancyClosedEventHandler, not this one. Or create a separate out of process event handler which initiates the comm sending stuff

src/Shared/Recruit.Vacancies.Client/Application/EventHandlers/NotifyOnVacancyActionsEventHandler.cs
dev4valtech: Try catch should be close to where we make the actual infrastructure call rather than interim classes unless we know how to treat the error situation.

PR #763, opened 2019-08-21
ER-1354_update banned phrases.

shomavg:


shomavg:
> Can you make the formatting the same as the existing phrases please. (I think they are using spaces rather than a tab)

👍

PR #771, opened 2019-08-23
ER-1406 - Added content for employer vacancies closed due to blocked provider

chrisjensenuk:
Added some content to the Manage vacancy screen for closed vacancies.

Also added fix so:

> 1. if the vacancy to be closed is one where an employer has given permission (so the provider created it) then it is transferred to the employer. closure reason = transferred by QA
> 2. if the vacancy to be closed is NOT one where an employer has given permission (they created it themselves and chose the provider) then it is just closed. closure reason = blocked by QA

### DIFF
--- a/src/Jobs/Recruit.Vacancies.Jobs/DomainEvents/Handlers/Vacancy/ProviderBlockedOnVacancyDomainEventHandler.cs
+++ b/src/Jobs/Recruit.Vacancies.Jobs/DomainEvents/Handlers/Vacancy/ProviderBlockedOnVacancyDomainEventHandler.cs
@@ -31,7 +31,9 @@ namespace Esfa.Recruit.Vacancies.Jobs.DomainEvents.Handlers.Vacancy
 
             var vacancy = await _vacancyRepository.GetVacancyAsync(eventData.VacancyId);
 
-            if (vacancy.OwnerType == Entities.OwnerType.Provider)
+            var isTransfer = vacancy.OwnerType == Entities.OwnerType.Provider;
+
+            if (isTransfer)
             {
                 await _messaging.SendCommandAsync(
                     new TransferProviderVacancyCommand(
@@ -49,7 +51,9 @@ namespace Esfa.Recruit.Vacancies.Jobs.DomainEvents.Handlers.Vacancy
 
             if (vacancy.Status == Entities.VacancyStatus.Live)
             {
-                await _messaging.SendCommandAsync(new CloseVacancyCommand(vacancy.Id, eventData.QaVacancyUser, Entities.ClosureReason.BlockedByQa));
+                var closureReason = isTransfer ? Entities.ClosureReason.TransferredByQa : Entities.ClosureReason.BlockedByQa;
+
+                await _messaging.SendCommandAsync(new CloseVacancyCommand(vacancy.Id, eventData.QaVacancyUser, closureReason));
             }
         }
     }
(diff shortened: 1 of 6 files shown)


src/Jobs/Recruit.Vacancies.Jobs/DomainEvents/Handlers/Vacancy/ProviderBlockedOnVacancyDomainEventHandler.cs
```diff
@@ -31,7 +31,9 @@ public async Task HandleAsync(string eventPayload)
 
             var vacancy = await _vacancyRepository.GetVacancyAsync(eventData.VacancyId);
 
-            if (vacancy.OwnerType == Entities.OwnerType.Provider)
+            var isTransfer = vacancy.OwnerType == Entities.OwnerType.Provider;
```
sachpatel: `hasToBeTransferred` maybe

src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/ApproveVacancyReviewCommandHandler.cs
```diff
@@ -88,7 +88,7 @@ private ClosureReason GetClosureReasonBasedOnTransferInfo(TransferInfo transferI
             }
             else
             {
-                return ClosureReason.BlockedByQa;
+                return ClosureReason.TransferredByQa;
```
sachpatel: This could still be the result of BlockedByQa. Think this method needs to be rewritten and not exclusively depend on TransferReason alone to work out what the closure reason is. The particular case in this is that there is an open `UnderReview` vacancy review for an employer created vacancy for a provider which has been blocked by QA. The employer owned vacancy wouldn't have been transferred, just closed in this case.

src/Jobs/Recruit.Vacancies.Jobs/DomainEvents/Handlers/Vacancy/ProviderBlockedOnVacancyDomainEventHandler.cs
chrisjensenuk: 👍

src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/ApproveVacancyReviewCommandHandler.cs
chrisjensenuk: 👍 Good spot

src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/ApproveVacancyReviewCommandHandler.cs
```diff
@@ -101,13 +103,27 @@ private void Validate(VacancyReview review)
...
         }
 
-        private bool GetHasVacancyBeenTransferredSinceReviewWasCreated(VacancyReview review, Vacancy vacancy)
+        private bool HasVacancyBeenTransferredSinceReviewWasCreated(VacancyReview review, Vacancy vacancy)
+        {
+            return review.VacancySnapshot.TransferInfo == null && vacancy.TransferInfo != null;
+        }
+
+        private async Task<bool> HasProviderBeenBlockedSinceReviewWasCreatedAsync(Vacancy vacancy)
+        {
```
sachpatel: This looks fine, my only slight concern is that if the provider is blocked/unblocked quickly but the source vacancy has been transferred, we are not looking at the source vacancy transfer info reason or state to work out the closure reason.

chrisjensenuk: For transferred vacancies we do look at the vacancy TransferInfo to determine the closure reason.  See `HasVacancyBeenTransferredSinceReviewWasCreated`

This handles non-transferred vacancies (Employer vacancies) for providers that have been blocked since the review was created. I think this is the only way we can determine this.

PR #772, opened 2019-08-23
ER-1391 Remove BlockedEmployers list

mrpmorris:
This PR removes the BlockedEmployers list in referenceData - and instead uses API to determine if an employer should have access to the site or not.

Note: A separate ticket will be created to update config files and change release templates for *after* a version with these changes has been released into PROD.

- Abstract EoiAgreementCookieWriter using an interface
- Add filter unit tests
- Rename methods for clarity
- Common code to read cookie for current user + employer ID
- Refactor EOI code
- Refactor levy code
- Remove blocked employers code
- Rename filter + remove feature toggle
- Remove blocked-employers scripts
- Use Lax instead of SameSite.Strict cookies in Dev mode
- Delete cookies when signing out


src/Shared/Recruit.Vacancies.Client/Infrastructure/Client/VacancyClient.cs
```diff
@@ -244,9 +245,9 @@ public Task<IEnumerable<string>> GetEmployerIdentifiersAsync(string userId)
             return _employerAccountProvider.GetEmployerIdentifiersAsync(userId);
         }
 
-        public Task<bool> GetEmployerEOIAsync(string employeraccountId)
+        public Task<AccountDetailViewModel> GetEmployerAccountDetailsAsync(string employeraccountId)
```
sachpatel: `AccountDetailViewModel` is the wrong name from the type declared within the EmployerAccount client and types packages. It isn't a ViewModel, as such and for other reasons we should not let these types bleed into other bits of our codebase outside f the Provider class. Can you make the reference to type with the name `AccountDetailViewModel` only be contained within the `EmployerAccountProvider` class and map the object into a type we declare which is named `EmployerAccount` please. There is a bit of a mix and match and a need to refactor further particularly the `LegalEntityViewModel` which is a similar issue. However the best thing we can do now is not add more.

src/Employer/Employer.Web/Services/IEoiAgreementCookieWriter.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Microsoft.AspNetCore.Http;
+
+namespace Esfa.Recruit.Employer.Web.Services
+{
+    public interface IEoiAgreementCookieWriter
+    {
+        string GetCookieFromRequest(HttpContext context);
+        void WriteCookie(HttpResponse response, string userId, string employerAccountId, bool eoiIsPresent);
```
sachpatel: `isEoiPresent` maybe

src/Employer/Employer.Web/Filters/EmployerAccessCheckFilter.cs
```diff
@@ -0,0 +1,264 @@
...
+        {
+            bool hasEoi = false;
+            if (GetCookieValueForUserAndEmployer(
+                context.HttpContext,
+                userId: userId,
+                employerAccountId: employerAccountId,
+                readValue: _eoiCookieWriter.GetCookieFromRequest,
+                result: out string hasEoiAsString))
+            {
+                hasEoi = bool.Parse(hasEoiAsString);
```
sachpatel: 👍

src/Employer/Employer.Web/Filters/EmployerAccessCheckFilter.cs
```diff
@@ -0,0 +1,264 @@
...
+            }
+            else
+            {
+                hasEoi = await GetEmployerHasEoi(employerAccountId);
+                _eoiCookieWriter.WriteCookie(context.HttpContext.Response, userId, employerAccountId, hasEoi);
+
+                // Why redirect to the dashboard?
```
sachpatel: Do we need this comment?

src/Employer/Employer.Web/Filters/EmployerAccessCheckFilter.cs
```diff
@@ -0,0 +1,264 @@
...
+                    employerAccountId: employerAccountId,
+                    userId: userId);
+                _levyCookieWriter.WriteCookie(context.HttpContext.Response, userId, employerAccountId, hasLevyDeclaration);
+            }
+
+            // No Levy declaration = not handled
+            if (!hasLevyDeclaration)
+                return false;
+
+            // Don't allow Levy pages if Levy is already declared
```
sachpatel: If they have Levy declared then we should allow access to Levy pages right

src/Employer/Employer.Web/Services/EmployerAccountTypeCookieWriter.cs
```diff
@@ -0,0 +1,59 @@
...
+        private readonly IDataProtector _dataProtector;
+        private readonly ILogger<EmployerAccountTypeCookieWriter> _logger;
+
+        public EmployerAccountTypeCookieWriter(
+            IHostingEnvironment hostingEnvironment,
+            IDataProtectionProvider dataProtectionProvider,
+            ILogger<EmployerAccountTypeCookieWriter> logger)
+        {
+            _hostingEnvironment = hostingEnvironment;
+            _dataProtector = dataProtectionProvider.CreateProtector(DataProtectionPurposes.EmployerAccountTypeCookie);
```
sachpatel: Do you think we should maybe declare this as a singleton or passed in from the Ioc for creating `EmployerAccountTypeCookieWriter`?

src/Employer/Employer.Web/Services/EoiAgreementCookieWriter.cs
```diff
@@ -41,9 +41,9 @@ public string GetCookieFromRequest(HttpContext context)
             return content;
         }
 
-        public void WriteCookie(HttpResponse response, string userId, string employerAccountId)
+        public void WriteCookie(HttpResponse response, string userId, string employerAccountId, bool eoiIsPresent)
```
sachpatel: `isEoiPresent` or `hasEoi` maybe

src/Employer/Employer.Web/Filters/EmployerAccessCheckFilter.cs
```diff
@@ -0,0 +1,264 @@
...
+
+    /// <remarks>
+    /// Using the employer code in the URL this filter will ensure the employer
+    /// meets at least one the following criteria
+    /// 1: Request is for a page that doesn't require special access
+    /// 2: Has a signed levy declaration
+    /// 3: Has an Expression Of Interest
+    /// 
+    /// Expected results:
+    /// A: Employers with a levy declaration cannot go to any levy pages - redirect to dashboard
```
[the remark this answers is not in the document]
mrpmorris: No. The rules are that if you have signed the levy you cannot go back to the levy page - that's what the code did :)

src/Employer/Employer.Web/Filters/EmployerAccessCheckFilter.cs
```diff
@@ -0,0 +1,264 @@
...
+    /// C: Non-levy employers with no EOI should be blocked
+    /// </remarks>
+    public class EmployerAccessCheckFilter : IAsyncActionFilter, IOrderedFilter
+    {
+        private readonly ILogger<EmployerAccessCheckFilter> _logger;
+        private readonly ILevyDeclarationCookieWriter _levyCookieWriter;
+        private readonly IEoiAgreementCookieWriter _eoiCookieWriter;
+        private readonly IEmployerAccountTypeCookieWriter _employerAccountTypeCookieWriter;
+        private readonly IRecruitVacancyClient _recruitVacancyClient;
+        private readonly string[] _allowedAccountTypes = new string[] { "levy", "nonlevy" };
```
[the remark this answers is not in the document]
mrpmorris: We are supposed to allow non-levy if they have an EOI

src/Employer/Employer.Web/Filters/EmployerAccessCheckFilter.cs
mrpmorris: Not according to the original code, no

src/Employer/Employer.Web/Services/EmployerAccountTypeCookieWriter.cs
mrpmorris: I don't understand. This class is only used via dependency injection by its interface.

src/Employer/Employer.Web/Filters/EmployerAccessCheckFilter.cs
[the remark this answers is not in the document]
sachpatel: I misunderstood and realise it is something different to what I thought it was.

src/Employer/Employer.Web/Filters/EmployerAccessCheckFilter.cs
mrpmorris: Not really, but it made no sense to me and I wanted it on record :)

src/Shared/Recruit.Vacancies.Client/Infrastructure/Client/VacancyClient.cs
mrpmorris: Done

PR #774, opened 2019-08-27
ER-1359 - wages and pay. Employer.Web

chrisjensenuk:
Updated content on wages page to provider guidance on minimum wages.

### DIFF
--- a/src/Employer/Employer.Web/ViewModels/Part1/Wage/WageViewModel.cs
+++ b/src/Employer/Employer.Web/ViewModels/Part1/Wage/WageViewModel.cs
@@ -15,6 +15,14 @@ namespace Esfa.Recruit.Employer.Web.ViewModels.Part1.Wage
 
         public string WageAdditionalInformation { get; set; }
 
+        public string MinimumWageStartFrom { get; set; }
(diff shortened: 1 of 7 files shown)


src/Employer/Employer.Web/ViewModels/Part1/Wage/WageViewModel.cs
```diff
@@ -15,6 +15,14 @@ public class WageViewModel
 
         public string WageAdditionalInformation { get; set; }
 
+        public string MinimumWageStartFrom { get; set; }
+        public string NationalMinimumWageLowerBound { get; set; }
+        public string NationalMinimumWageUpperBound { get; set; }
+        public string NationalMinimumWageYearly { get; set; }
+        public string ApprenticeshipMinimumWage { get; set; }
```
dev4valtech: May be add Hourly to the name to make it obvious and consistent with "Yearly". Also same with lower and upper bound fields.

chrisjensenuk: 👍 I'll do this in a future PR

PR #778, opened 2019-08-28
ER-1485

sachpatel:
Exclude vacancy reviews from QA search if the vacancy review has been closed due to vacancy transfer

### DIFF
--- a/src/QA/QA.Web/Orchestrators/DashboardOrchestrator.cs
+++ b/src/QA/QA.Web/Orchestrators/DashboardOrchestrator.cs
@@ -44,14 +44,17 @@ namespace Esfa.Recruit.Qa.Web.Orchestrators
                 vm.InProgressVacancies = inProgressVacanciesVm.ToList();
             }
 
-            if(string.IsNullOrEmpty(searchTerm)) return vm;
+            if (string.IsNullOrEmpty(searchTerm)) return vm;
 
             vm.LastSearchTerm = searchTerm;
             var matchedVacancyReview = await _vacancyClient.GetSearchResultAsync(searchTerm);
-            var searchResults = matchedVacancyReview != null ? new List<VacancyReview>(){ matchedVacancyReview } : new List<VacancyReview>();
-            var searchResultsVmTasks = searchResults.Select(v => MapToViewModelAsync(v, vacancyUser));
-            var searchResultsVm = await Task.WhenAll(searchResultsVmTasks);
-            vm.SearchResults = searchResultsVm.ToList();
+
+            if (matchedVacancyReview != null)
+            {
+                var matchVacancyReviewSearchVm = await MapToViewModelAsync(matchedVacancyReview, vacancyUser);
+                vm.SearchResult = matchVacancyReviewSearchVm;
+            }
+
             return vm;
         }
 
--- a/src/QA/QA.Web/Orchestrators/DashboardOrchestrator.cs
+++ b/src/QA/QA.Web/Orchestrators/DashboardOrchestrator.cs
@@ -10,6 +10,7 @@ using Esfa.Recruit.Vacancies.Client.Domain.Entities;
 using Esfa.Recruit.Vacancies.Client.Infrastructure.Client;
 using Esfa.Recruit.Vacancies.Client.Infrastructure.QueryStore.Projections.QA;
 using Esfa.Recruit.Vacancies.Client.Domain.Extensions;
+using System.Collections.Generic;
 
 namespace Esfa.Recruit.Qa.Web.Orchestrators
 {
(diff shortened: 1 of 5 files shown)


src/Shared/Recruit.Vacancies.Client/Infrastructure/Client/QaVacancyClient.cs
```diff
@@ -81,20 +81,29 @@ public async Task<QaDashboard> GetDashboardAsync()
             return dashboard;
         }
 
-        public async Task<List<VacancyReview>> GetSearchResultsAsync(string searchTerm)
+        public async Task<VacancyReview> GetSearchResultAsync(string searchTerm)
         {
             //The return type is a list to allow adding results from other searches to be implemented in future
             var result = new List<VacancyReview>();
```
chrisjensenuk: Kill this and the comment above.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Client/QaVacancyClient.cs
```diff
@@ -81,20 +81,29 @@ public async Task<QaDashboard> GetDashboardAsync()
...
-            if (!TryGetVacancyReference(searchTerm, out var vacancyReference)) return result;
+            if (TryGetVacancyReference(searchTerm, out var vacancyReference) == false) return null;
 
             var review = await _vacancyReviewQuery.GetLatestReviewByReferenceAsync(vacancyReference);
-            if (review != null && review.Status != ReviewStatus.New) result.Add(review);
 
-            return result;
+            if (review != null)
+            {
+                var canDisplayReview = review.Status != ReviewStatus.New && (review.ManualOutcome == null || review.ManualOutcome != ManualQaOutcome.Transferred);
```
chrisjensenuk: Do we need `review.ManualOutcome == null ||` ?

src/QA/QA.Web/Orchestrators/DashboardOrchestrator.cs
```diff
@@ -46,8 +47,9 @@ public async Task<DashboardViewModel> GetDashboardViewModelAsync(string searchTe
             if(string.IsNullOrEmpty(searchTerm)) return vm;
 
             vm.LastSearchTerm = searchTerm;
-            var searchResults = await _vacancyClient.GetSearchResultsAsync(searchTerm);
-            var searchResultsVmTasks = searchResults.Select(v => MapToViewModelAsync(v, vacancyUser)).ToList();
+            var matchedVacancyReview = await _vacancyClient.GetSearchResultAsync(searchTerm);
+            var searchResults = matchedVacancyReview != null ? new List<VacancyReview>(){ matchedVacancyReview } : new List<VacancyReview>();
```
chrisjensenuk: If we're changing the GetSearchResultAsync to return a single result we may as well chase it through to the ViewModel and the View.  So change `vm.SearchResults` to `vm.SearchResult`

src/Shared/Recruit.Vacancies.Client/Infrastructure/Client/QaVacancyClient.cs
```diff
@@ -81,20 +81,29 @@ public async Task<QaDashboard> GetDashboardAsync()
...
         }
 
-        public async Task<List<VacancyReview>> GetSearchResultsAsync(string searchTerm)
+        public async Task<VacancyReview> GetSearchResultAsync(string searchTerm)
         {
             //The return type is a list to allow adding results from other searches to be implemented in future
             var result = new List<VacancyReview>();
 
-            if (!TryGetVacancyReference(searchTerm, out var vacancyReference)) return result;
+            if (TryGetVacancyReference(searchTerm, out var vacancyReference) == false) return null;
```
dev4valtech: can you return the result here which is an empty list. I don't like returning nulls, which means the consumer has to do more.

dev4valtech: can you return the result here which is an empty list. I don't like returning nulls, which means the consumer has to do more.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Client/QaVacancyClient.cs
sachpatel: 👍

src/Shared/Recruit.Vacancies.Client/Infrastructure/Client/QaVacancyClient.cs
sachpatel: Yeah I believe so because ManualOutcome is nullable

src/QA/QA.Web/Orchestrators/DashboardOrchestrator.cs
sachpatel: Will look into this. I was trying to avoid changing too much. The client method sounded wrong to what it was doing.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Client/QaVacancyClient.cs
sachpatel: The client method sounded wrong to what it was doing. Which is why I have changed it to null. In the dashboard orchestrator is where the work should be done in my opinion.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Client/QaVacancyClient.cs
dev4valtech: don't kill the comment... it is important as otherwise it does not make sense.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Client/QaVacancyClient.cs
dev4valtech: I see you have removed the result from being a list, so this is fine.

PR #779, opened 2019-08-28
ER-1489 - wage and pay validations. Employer.Web

chrisjensenuk:
Added the concept of `Soft Validations`  these get run on the GET of the vacancy preview.

Refactored `_ValidationSummaryPartial.cshtml ` to pull out the generation of the error list into a separate partial `_ErrorListItemPartial.cshtml` so we can reuse when building the soft validations summary.

Refactored `WageOrchestrator.GetMinimumWageYearlyText` so directly uses `WagePresenter` without having to create a temporary `Wage`object.

### DIFF
--- a/src/Employer/Employer.Web/Orchestrators/VacancyPreviewOrchestrator.cs
+++ b/src/Employer/Employer.Web/Orchestrators/VacancyPreviewOrchestrator.cs
@@ -21,6 +21,8 @@ namespace Esfa.Recruit.Employer.Web.Orchestrators
     public class VacancyPreviewOrchestrator : EntityValidatingOrchestrator<Vacancy, VacancyPreviewViewModel>
     {
         private const VacancyRuleSet ValidationRules = VacancyRuleSet.All;
+        private const VacancyRuleSet SoftValidationRules = VacancyRuleSet.MinimumWage;
+
         private readonly IEmployerVacancyClient _client;
         private readonly IRecruitVacancyClient _vacancyClient;
         private readonly DisplayVacancyViewModelMapper _vacancyDisplayMapper;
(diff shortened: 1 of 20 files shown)


src/Employer/Employer.Web/Orchestrators/VacancyPreviewOrchestrator.cs
```diff
@@ -115,22 +119,31 @@ private async Task<SubmitVacancyResponse> SubmitActionAsync(Vacancy vacancy, Vac
...
             foreach (var error in errors.Where(e => e.PropertyName.StartsWith(nameof(Vacancy.Qualifications))))
             {
                 error.PropertyName = nameof(VacancyPreviewViewModel.Qualifications);
             }
         }
 
+        private EntityValidationResult GetSoftValidationErrors(Vacancy vacancy)
+        {
+            var result = _vacancyClient.Validate(vacancy, SoftValidationRules);
+            FlattenErrors(result.Errors);
```
sachpatel: Do we need to run FlattenErrors here when you are only validating on MinimumWage and Flatten Errors is really only specific for Qualifications errors?

src/Employer/Employer.Web/ViewModels/VacancyPreview/VacancyPreviewViewModel.cs
```diff
@@ -32,7 +33,10 @@ public class VacancyPreviewViewModel : DisplayVacancyViewModel
         public VacancyPreviewSectionState WageTextSectionState { get; internal set; }
         public VacancyPreviewSectionState DescriptionsSectionState { get; internal set; }
         public VacancyPreviewSectionState WorkingWeekSectionState { get; internal set; }
-        
+
+        public EntityValidationResult SoftValidationErrors { get; internal set; }
+        public bool HideValidationSummary { get; internal set; }
```
sachpatel: `CanHideValidationSummary` please

src/Employer/Employer.Web/Orchestrators/VacancyPreviewOrchestrator.cs
chrisjensenuk: This was my effort to keep the mapping of vacancy to ModelState the same between the Get and the POST.  I'll move this into a common mapping function to be more explicit. 👍

src/Employer/Employer.Web/ViewModels/VacancyPreview/VacancyPreviewViewModel.cs
chrisjensenuk: 👍

PR #785, opened 2019-08-30
ER-1485 -Removing Withdrawn as a possible ManualOutcome

sachpatel:


sachpatel:
@dev4valtech just had a word with Mark and it should be `Blocked` as you have said rather than `Withdrawn` I will raise another PR to correct this.

PR #789, opened 2019-09-02
ER-1345 - Dates validation. Provider.Web

chrisjensenuk:
Added extra validation to the Dates page to handle expiring training programmes.  Also added this check to soft validations on vacancy preview.

There is also different content we show depending on whether the training has an expiry date or whether there is an error is displayed due to the training expiry date.

### DIFF
--- a/src/Provider/Provider.Web/Controllers/Part1/DatesController.cs
+++ b/src/Provider/Provider.Web/Controllers/Part1/DatesController.cs
@@ -1,3 +1,4 @@
+using System.Linq;
 using System.Threading.Tasks;
 using Esfa.Recruit.Provider.Web.Configuration.Routing;
 using Esfa.Recruit.Provider.Web.Extensions;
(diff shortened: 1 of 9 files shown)


src/Provider/Provider.Web/Controllers/Part1/DatesController.cs
```diff
@@ -39,6 +41,7 @@ public async Task<IActionResult> Dates(DatesEditModel m, [FromQuery] bool wizard
             if (!ModelState.IsValid)
             {
                 var vm = await _orchestrator.GetDatesViewModelAsync(m);
+                vm.CanShowTrainingErrorHint = response.Errors.Errors.Any(e => e.ErrorCode == ErrorCodes.TrainingExpiryDate);
```
mrpmorris: I'd call it "ShouldShowTrainingErrorHint". This indicates that we want it displayed, rather than it is permitted to show it in the circumstances that we wish to show it.

PR #792, opened 2019-09-03
ER-1340 - Dates validation. Empoyer.Web

chrisjensenuk:
Added extra validation to the Dates page to handle expiring training programmes. Also added this check to soft validations on vacancy preview.

There is also different content we show depending on whether the training has an expiry date or whether there is an error is displayed due to the training expiry date.

### DIFF
--- a/src/Employer/Employer.Web/Orchestrators/Part1/DatesOrchestrator.cs
+++ b/src/Employer/Employer.Web/Orchestrators/Part1/DatesOrchestrator.cs
@@ -9,6 +9,7 @@ using Esfa.Recruit.Shared.Web.Services;
 using Esfa.Recruit.Vacancies.Client.Application.Providers;
 using Esfa.Recruit.Vacancies.Client.Application.Validation;
 using Esfa.Recruit.Vacancies.Client.Domain.Entities;
+using Esfa.Recruit.Vacancies.Client.Domain.Extensions;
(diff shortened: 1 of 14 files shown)


src/Employer/Employer.Web/Orchestrators/Part1/DatesOrchestrator.cs
```diff
@@ -98,6 +103,20 @@ public async Task<OrchestratorResponse> PostDatesEditModelAsync(DatesEditModel m
...
+        {
+            if (string.IsNullOrEmpty(vacancy.ProgrammeId))
+                return;
+
+            var programme = await _apprenticeshipProgrammeProvider.GetApprenticeshipProgrammeAsync(vacancy.ProgrammeId);
+
+            if (programme?.EffectiveTo.HasValue == true)
+            {
+                vm.TrainingDescription = $"{programme.Title}, Level:{(int)programme.Level}";
+                vm.TrainingEffectiveToDate = programme.EffectiveTo.Value.AsGdsDate();
```
sachpatel: What if EffectiveTo is null? this will go bang I think. `AsGdsDate` does not check

chrisjensenuk: 3 lines above there is a null check.

PR #795, opened 2019-09-04
ER-1446 - Provider permission check on submit

chrisjensenuk:
We need to check the provider permission upon vacancy submit. I've decided to do this check as part of the vacancy validation as we already do other provider checks (blocked & roatp) here too.

- I've added this permission check to the vacancy validation. It only runs for provider owned vacancies
- If the provider does not have permission they the user is redirected to the existing missing permissions error page.
- I've added a new function to the `ProviderRelationshipService` called `GetProviderPermissionsAsync` this checks that the provider has permissions for the specified employer account and legal entity.

### DIFF
--- a/src/Provider/Provider.Web/Orchestrators/ManageNotificationsOrchestrator.cs
+++ b/src/Provider/Provider.Web/Orchestrators/ManageNotificationsOrchestrator.cs
@@ -44,8 +44,6 @@ namespace Esfa.Recruit.Provider.Web.Orchestrators
                 return new OrchestratorResponse(_notificationTypeIsRequiredForTheFirstTime);
             }
 
-            if (editModel.IsApplicationSubmittedSelected) editModel.NotificationFrequency = NotificationFrequency.Immediately;
-
             var preferences = GetDomainModel(editModel, vacancyUser.UserId);
 
             return await ValidateAndExecute(


src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/ProviderRelationship/IProviderRelationshipsService.cs
```diff
@@ -7,6 +7,7 @@ namespace Esfa.Recruit.Vacancies.Client.Infrastructure.Services.ProviderRelation
     public interface IProviderRelationshipsService
     {
         Task<IEnumerable<EmployerInfo>> GetLegalEntitiesForProviderAsync(long ukprn);
+        Task<bool> ProviderHasPermissionAsync(long ukprn, string accountPublicHashedId, long legalEntityId);
```
dev4valtech: Consider renaming to `HasEmployerPermittedProviderAsync` or `HasProviderGotEmployersPermission`

src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/ProviderRelationship/ProviderRelationshipsService.cs
```diff
@@ -32,7 +32,39 @@ public class ProviderRelationshipsService : IProviderRelationshipsService
...
+                .ToList();
+
+            if (permittedLegalEntities == null || permittedLegalEntities.Any() == false)
+                return false;
+
+            var accountId = await _employerAccountProvider.GetEmployerAccountPublicHashedIdAsync(permittedLegalEntities.First().AccountId);
+            var allLegalEntities = (await _employerAccountProvider.GetLegalEntitiesConnectedToAccountAsync(accountId)).ToList();
+
+            var hasPermission = permittedLegalEntities.Join(allLegalEntities,
+                    ple => ple.AccountLegalEntityPublicHashedId,
```
dev4valtech: pale ale 😸 (ignore)

src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/ProviderRelationship/ProviderRelationshipsService.cs
```diff
@@ -44,10 +76,11 @@ public async Task<IEnumerable<EmployerInfo>> GetLegalEntitiesForProviderAsync(lo
                     if (!response.IsSuccessStatusCode)
                     {
                         _logger.LogError("An invalid response received when trying to get provider relationships");
-                        return new EmployerInfo[]{};
-                    }                
-                    var content  = await response.Content.ReadAsStringAsync();
-                    providerPermissions = JsonConvert.DeserializeObject<ProviderPermissions>(content);
+                        return null;
```
dev4valtech: I am not fan of returning nulls as far as possible, especially when the return value is a list. I saw where this is being consumed and we can easily check for empty list in both the cases.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/ProviderRelationship/ProviderRelationshipsService.cs
```diff
@@ -32,7 +32,39 @@ public class ProviderRelationshipsService : IProviderRelationshipsService
...
+
+        public async Task<bool> ProviderHasPermissionAsync(long ukprn, string accountPublicHashedId, long legalEntityId)
+        {
+            var providerPermissions = await GetProviderPermissionsAsync(ukprn);
+
+            var permittedLegalEntities = providerPermissions?.AccountProviderLegalEntities
+                .Where(l => l.AccountPublicHashedId == accountPublicHashedId)
+                .ToList();
+
+            if (permittedLegalEntities == null || permittedLegalEntities.Any() == false)
```
dev4valtech: (see my comment on `GetProviderPermissionsAsync` first) I think if `GetProviderPermissionsAsync` is updated to return empty list, then the check for `.Any()` can be moved up before populating `permittedLegalEntities`.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/ProviderRelationship/ProviderRelationshipsService.cs
```diff
@@ -32,7 +32,39 @@ public class ProviderRelationshipsService : IProviderRelationshipsService
 
         public async Task<IEnumerable<EmployerInfo>> GetLegalEntitiesForProviderAsync(long ukprn)
         {
-            ProviderPermissions providerPermissions = null;
+            var providerPermissions = await GetProviderPermissionsAsync(ukprn);
```
dev4valtech: (see my comment on `GetProviderPermissionsAsync` first) I think if `GetProviderPermissionsAsync` is updated to return empty list, then the check for `.Any()` will work here.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/ProviderRelationship/ProviderRelationshipsService.cs
```diff
@@ -58,7 +91,7 @@ public async Task<IEnumerable<EmployerInfo>> GetLegalEntitiesForProviderAsync(lo
                     _logger.LogError(ex, $"Couldn't deserialise {nameof(ProviderPermissions)}.", null);
                 }
 
-                return await GetEmployerInfosAsync(providerPermissions);
+                return null;
```
dev4valtech: throw exception here perhaps as this is not a condition we expect and consuming code will have to handle this.

src/Employer/Esfa.Recruit.Azure.Resources/azuredeploy.json
```diff
@@ -755,6 +755,26 @@
               {
                 "name": "PasAccountApiConfiguration:Tenant",
                 "value": "[parameters('PasAccountApiConfiguration').Tenant]"
+              },
```
sachpatel: Could you update the developer setup page on Confluence to include this in user secrets as well please for Provider.Web

src/Shared/UnitTests/Vacancies.Client/Application/VacancyValidation/SingleField/TrainingProviderUkprnValidationTests.cs
```diff
@@ -115,5 +117,60 @@ public void TrainingProviderUkprnMustBe8Digits(long? ukprn)
...
+
+            var vacancy = new Vacancy
+            {
+                OwnerType = OwnerType.Provider,
+                TrainingProvider = new TrainingProvider { Ukprn = ukprn },
+                EmployerAccountId = employerAccountId,
+                LegalEntityId = legalEntityId
+            };
+
+            MockTrainingProviderSummaryProvider.Setup(p => p.GetAsync(12345678)).ReturnsAsync(new TrainingProviderSummary());
```
sachpatel: Can you use the ukprn const in this call setup please like you have in the test method below.

src/Shared/Recruit.Vacancies.Client/Application/Validation/Fluent/CustomValidators/VacancyValidators/VacancyFluentValidationExtensions.cs
```diff
@@ -116,6 +117,27 @@ await trainingProviderSummaryProvider.GetAsync(trainingProvider.Ukprn.Value) !=
             });
         }
 
+        internal static IRuleBuilderInitial<Vacancy, Vacancy> TrainingProviderVacancyMustHaveEmployerPermission(this IRuleBuilder<Vacancy, Vacancy> ruleBuilder, IProviderRelationshipsService providerRelationshipService)
```
sachpatel: In this class there is a method called `TrainingProviderMustExistInRoatp` and a call within a `if` condition like `await trainingProviderSummaryProvider.GetAsync(trainingProvider.Ukprn.Value) != null` where I believe the await call might not be evaluated before the `!= null` check. could you stick brackets around the await portion of this line. Otherwise I think it is checking != against a Task rather than the actual result of the async call.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/ProviderRelationship/IProviderRelationshipsService.cs
chrisjensenuk: 👍 I'll go with `HasProviderGotEmployersPermissionAsync`

src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/ProviderRelationship/ProviderRelationshipsService.cs
chrisjensenuk: 🍺

src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/ProviderRelationship/ProviderRelationshipsService.cs
chrisjensenuk: 👍   I'll return a `new ProviderPermissions { AccountProviderLegalEntities = Enumerable.Empty<LegalEntityDto>() };` then we won't need to do a null check

src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/ProviderRelationship/ProviderRelationshipsService.cs
chrisjensenuk: I can remove the null check but I still need to keep this check here as the permissions get filtered to the employer account so could well be empty at this point.

src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/ProviderRelationship/ProviderRelationshipsService.cs
chrisjensenuk: 👍 I can remove the below line below entirely

src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/ProviderRelationship/ProviderRelationshipsService.cs
chrisjensenuk: I have refactored so all non-happy paths log an error and return an empty list of permissions.

src/Employer/Esfa.Recruit.Azure.Resources/azuredeploy.json
chrisjensenuk: 👍 I've done this now.

src/Shared/UnitTests/Vacancies.Client/Application/VacancyValidation/SingleField/TrainingProviderUkprnValidationTests.cs
chrisjensenuk: 👍

src/Shared/Recruit.Vacancies.Client/Application/Validation/Fluent/CustomValidators/VacancyValidators/VacancyFluentValidationExtensions.cs
chrisjensenuk: 👍

PR #815, opened 2019-09-10
ER-1486 - Transfer without LegalEntityId

chrisjensenuk:
We floated the idea of removing the queue trigger handler but this is contentious and I didn't need to do this refactoring to get this ticket done. So I have deferred for now and I'll add an item in the tech debt to re-review.  

I'm on the fence about removing this handler.
PRO: Simplifies the code base by removing a handler and queue
CON: For testing it requires code changes to setup the Spike handler.


dev4valtech:
This PR is over ridden by https://github.com/SkillsFundingAgency/das-recruit/pull/847

PR #816, opened 2019-09-10
ER-1433 display educationlevels +semver: minor

mrpmorris:


src/Shared/Recruit.Shared.Web/Helpers/EducationLevelNumberHelper.cs
```diff
@@ -1,8 +1,11 @@
-﻿namespace Esfa.Recruit.Shared.Web.Helpers
+﻿using Esfa.Recruit.Vacancies.Client.Domain.Entities;
+using MongoDB.Driver;
+
+namespace Esfa.Recruit.Shared.Web.Helpers
 {
     public static class EducationLevelNumberHelper
     {
-        public static string GetEducationLevelName(int? educationLevelNumber)
+        private static string GetEducationLevelName(int? educationLevelNumber)
```
dev4valtech: Could change the signature here... the overload method is just a bypass and not necessary

mrpmorris: I want that method to be private because it shouldn't really be used. We always need a value, and the ApprenticeshipLevel is the fallback.

PR #818, opened 2019-09-11
ER-1427 - Ignoring withdrawn applications on dashboard count

chrisjensenuk:
We're now ignoring withdrawn vacancies when calculating the count for New, Successful and Unsuccessful applications.

The bigger change is I've simplified `VacancySummaryAggQueryBuilder` so it just loads the entire aggregate pipeline from a string rather than composing it using a series of BsonDocuments and BsonArrays.  This makes it much easier (for me) to debug as you don't need to interpret it and can just cut-n-paste into RoboMongo.  Hope this refactor isn't too contentious. I took the same approach with the Provider reports.

### DIFF
--- a/src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/VacancySummariesProvider/VacancySummaryAggQueryBuilder.cs
+++ b/src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/VacancySummariesProvider/VacancySummaryAggQueryBuilder.cs
@@ -22,13 +22,18 @@ namespace Esfa.Recruit.Vacancies.Client.Infrastructure.Services.VacancySummaries
                 }
             },
             {
+                '$match': { 
+                    '$or': [
+                    { 'candidateApplicationReview.isWithdrawn': {'$exists' : false} }, 
+                    { 'candidateApplicationReview.isWithdrawn': false }] }
+            },
+            {
                 '$project': {
                     'vacancyGuid': '$_id',
                     'vacancyReference': 1,
                     'title': 1,
                     'status': 1,
                     'appStatus': '$candidateApplicationReview.status',
-                    'appIsWithdrawn': '$candidateApplicationReview.isWithdrawn',
                     'legalEntityId': 1,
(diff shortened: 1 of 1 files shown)


src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/VacancySummariesProvider/VacancySummaryAggQueryBuilder.cs
```diff
@@ -1,176 +1,166 @@
...
-                    { "transferInfoTransferredDate", 1 },
-                    { "transferInfoReason", 1 },
-                    { "trainingProviderName", 1 }
+                '$project': {
+                    'vacancyGuid': '$_id',
+                    'vacancyReference': 1,
+                    'title': 1,
+                    'status': 1,
+                    'appStatus': '$candidateApplicationReview.status',
+                    'appIsWithdrawn': '$candidateApplicationReview.isWithdrawn',
```
dev4valtech: Wouldn't it be easier if we have a match clause before this one to filter out withdrawn applications?

dev4valtech: > The bigger change is I've simplified `VacancySummaryAggQueryBuilder` so it just loads the entire aggregate pipeline from a string rather than composing it using a series of BsonDocuments and BsonArrays. 

Works for me. 👍

chrisjensenuk: @dev4valtech Yes it would and we could do this using a more recent version of mongo.  The Cosmos DB version doesn't allow you to add a match clause in the $lookup.

sachpatel: Why can't you use another match stage in the pipeline after to filter it out

PR #820, opened 2019-09-11
ER-1482 send vacancy closed notification on review approval +semver: …

dev4valtech:
…patch

A new requirement was added to the story. When a review is approved for the vacancy that has blocked provider, the vacancy is closed instead of making it to live. The requirement is to also notify employer about the closed vacancy. 

I found that the Communication.Core had dependency on Recuirt.Vacancies.Client which is weird as any of the Communication projects should not have dependency on any of the recruit projects. I have removed this and refactored code/tests accordingly. 

Since I had to create similar communication request, I moved this to a factory class. 

Also I have removed some of the Client class methods as part of cleanup.

### DIFF
--- a/src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/ApproveVacancyReviewCommandHandler.cs
+++ b/src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/ApproveVacancyReviewCommandHandler.cs
@@ -10,6 +10,9 @@ using Esfa.Recruit.Vacancies.Client.Domain.Events;
 using FluentValidation;
 using Microsoft.Extensions.Logging;
 using Esfa.Recruit.Vacancies.Client.Infrastructure.Services.Projections;
+using Communication.Types;
+using Esfa.Recruit.Vacancies.Client.Application.Communications;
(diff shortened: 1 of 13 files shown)


src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/ApproveVacancyReviewCommandHandler.cs
```diff
@@ -76,13 +83,21 @@ public async Task Handle(ApproveVacancyReviewCommand message, CancellationToken
...
                 await _dashboardService.ReBuildDashboardAsync(vacancy.EmployerAccountId);
                 return;
             }
 
             await PublishVacancyReviewApprovedEventAsync(message, review);    
         }
 
+        private Task SendNotificationToEmployerAsync(long ukprn, string employerAccountId)
+        {
+            var communicationRequest = CommunicationRequestFactory.GetProviderBlockedEmployerNotificationForLiveVacanciesRequest(ukprn, employerAccountId);
```
chrisjensenuk: `TryGetReasonToCloseVacancy` suggests the ClosureReason could be TransferredByEmployer, TransferredByQa or BlockedByQa.  Is this communication appropriate for each of these reasons?

PR #829, opened 2019-09-18
ER-1533_show resubmitted date title

shomavg:


src/QA/QA.Web/Mappings/ReviewMapper.cs
```diff
@@ -216,15 +218,13 @@ public async Task<ReviewViewModel> Map(VacancyReview review)
                 vm.FieldIdentifiers = await GetFieldIdentifiersViewModel(review);
                 vm.ReviewerComment = review.ManualQaComment;
                 vm.ReviewHistories = historiesVm;
-                vm.IsResubmission = historiesVm.HasHistories;
+                vm.IsResubmission = historiesVm.HasHistories && currentVacancyResult.Status == VacancyStatus.Submitted;
```
chrisjensenuk: If the 2nd review get rejected and the the QA views it would this say Submitted instead of Resubmitted which is not right.

shomavg: im using SubmissionCount instead.

PR #831, opened 2019-09-18
ER-1518 Implements house keeping job for query store collection +semv…

dev4valtech:
…er: minor

The timer trigger will raise message with a date time stamp, this will be used as the base date in the queue trigger. This will enable testing by dropping the message manually. Also there is a fall back to use today's date if the message doesn't have one. 

I had to convert based class to be non-abstract so I can query any view types. 

There are 5000+ EditVacancyInfo documents that qualify for deletion. I tested my code with 6000. I am using DeleteMany command, which only managed about 2500 after 3 attempts due to polly retry.  So it will require about 3 messages to begin with to clear the whole lot. But after that first run this should execute ideally.


dev4valtech:
> I would prefer if the config value you introduced was used in the appSettings config as I am sure you will be moving towards the storage config used by the rest of the service but I will leave that up to you. Haven't tested but this looks okay to me.

What was the purpose of Configuration collection, was the view to move these to global config as well?

PR #834, opened 2019-09-24
ER-1464_Show legal entity name

shomavg:


src/Employer/Employer.Web/Orchestrators/LegalEntityAgreementOrchestrator.cs
```diff
@@ -31,13 +31,21 @@ public class LegalEntityAgreementOrchestrator
                     _client, _vacancyClient, vrm, RouteNames.LegalEntityAgreement_SoftStop_Get);
 
             var legalEntityId = selectedLegalEntityId.HasValue ? selectedLegalEntityId.Value : vacancy.LegalEntityId;
-            
+
+            var legalEntityTask =
+                _legalEntityAgreementService.GetLegalEntityAsync(vrm.EmployerAccountId, legalEntityId);
+            var hasLegalEntityAgreementTask =
```
chrisjensenuk: As discussed - these two functions both get the legal entity from the database.  If we add a new function to `LegalEntityAgreementService`. E.g.
 `public async Task<bool> HasLegalEntityAgreementAsync(LegalEntity legalEntity)`
Then we'll only need to get the legalEntity once from the Db.

PR #849, opened 2019-10-18
Remove personal information from log entries - Recruitment

VasanthaKasirajan3008:


src/Jobs/Recruit.Vacancies.Jobs/Communication/CommunicationMessageDispatcher.cs
```diff
@@ -74,7 +74,7 @@ public async Task Run(CommunicationMessageIdentifier commMsgId)
 
         public async Task SendEmail(CommunicationMessage request)
         {
-            _logger.LogInformation($"Trying to send message of type {request.RequestType} to {request.Recipient.Email}");
+            _logger.LogInformation($"Trying to send message of type {request.RequestType}");
```
dev4valtech: This actually kills the purpose of the log. Instead can we trim the domain from the email so that this log entry makes any sense. You can create an extension method to trim everything after `@` symbol in the email.

src/Shared/Recruit.Vacancies.Client/Application/CommandHandlers/GeocodeVacancyCommandHandler.cs
```diff
@@ -50,7 +50,7 @@ public async Task Handle(GeocodeVacancyCommand message, CancellationToken cancel
 
         private Task<Geocode> GeocodePostcodeAsync(Vacancy vacancy)
         {
-            _logger.LogInformation("Attempting to geocode postcode:'{postcode}' for vacancyId:'{vacancyId}'", vacancy.EmployerLocation.Postcode, vacancy.Id);
+            _logger.LogInformation("Attempting to geocode for vacancyId:'{vacancyId}'", vacancy.Id);
```
dev4valtech: While we are printing postcode here... but it has no association with any individual... so postcode in itself shouldn't be an issue?

PR #856, opened 2019-12-12
CON-1184 populate account legal entity public hashed id on vacancy

dev4valtech:
There are no unit tests and unfortunately the implementation of Utility has made it very difficult. We should have a tech debt to refactor utility to implement an interface to enable unit testing.

### DIFF
--- a/src/Jobs/Recruit.Vacancies.Jobs/Triggers/QueueTriggers/DeleteStaleQueryStoreDocumentsQueueTrigger.cs
+++ b/src/Jobs/Recruit.Vacancies.Jobs/Triggers/QueueTriggers/DeleteStaleQueryStoreDocumentsQueueTrigger.cs
@@ -59,8 +59,7 @@ namespace Esfa.Recruit.Vacancies.Jobs.Triggers.QueueTriggers
 
                 var documentTypesToDelete = new [] 
                 {
-                    nameof(EditVacancyInfo), 
-                    nameof(ProviderEditVacancyInfo),
+                    "EditVacancyInfo", 
                     nameof(EmployerDashboard),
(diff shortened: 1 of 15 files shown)


src/Jobs/Recruit.Vacancies.Jobs/Triggers/QueueTriggers/DeleteStaleQueryStoreDocumentsQueueTrigger.cs
```diff
@@ -59,8 +59,7 @@ public async Task DeleteStaleQueryStoreDocumentsAsync([QueueTrigger(QueueNames.D
 
                 var documentTypesToDelete = new [] 
                 {
-                    nameof(EditVacancyInfo), 
-                    nameof(ProviderEditVacancyInfo),
+                    "EditVacancyInfo",
```
cofaulco: any reason for not continuing the theme of nameof() ?

dev4valtech: There was fault (harmless) in the original code. EmployerEditVacancyInfo (formerly EditVacancyInfo) and ProviderEditVacancyInfo both have viewType of `EditVacancyInfo` and hence `nameof(ProviderEditVacancyInfo)` was not adding any value. Since I renamed the `EditVacancyInfo` to `EmployerEditVacancyInfo` I am unable to use `nameof` as I want to match `EditVacancyInfo` and not `EmployerEditVacancyInfo` or `ProviderEditVacancyInfo`. Hope this makes sense.

PR #859, opened 2019-12-20
CON-1185 Fix connection issues

dev4valtech:
Cause:
To update the AccountLegalEntityPublicHashedId on all the existing vacancies, I was queuing each vacancy individually to be processed by a job that will fetch the Id from EAS API and update the vacancy. The job uses the API client which was registered in the IOC as a transient. Which meant that each instance of the job was injected with new client instance in turn creating a new connection.

Effect:
This resulted in error Only one usage of each socket address (protocol/network address/port) is normally permitted for some vacancies and messages ended in the poison queue.

Resolution:
Changing the IOC registration of the EAS API client to be a singleton to use only one outbound connection. There are chances that this can backfire, only testing will tell.

Update:
Above solution didn't work, now querying and updating the vacancies straight away using parallel.


dev4valtech:
Closing this one as the solution didn't work.

PR #862, opened 2020-01-15
Recruitment - "Cookies" banner

VasanthaKasirajan3008:


shomavg:
Hi Vas,
You will need to add the new configurations to the azuredeploy.json files. They are found in this location - ..\src\Employer\Esfa.Recruit.Azure.Resources. The new configs should be added to both azuredeploy.json and azuredeploy.parameters.json. Take an example of an existing config and make similar changes.

VasanthaKasirajan3008:
> Please make additional changes mentioned in the comment. Other than that it all looks good. Can you please confirm from Nav/Chris if similar changes are needed for the Provider and QA site?

-- Provider comes under PAS and QA is not under MA,but QA and recruit are in the same repository, Deen already flagged Chris on this.

> Hi Vas,
> You will need to add the new configurations to the azuredeploy.json files. They are found in this location - ..\src\Employer\Esfa.Recruit.Azure.Resources. The new configs should be added to both azuredeploy.json and azuredeploy.parameters.json. Take an example of an existing config and make similar changes.

I have done routing changes in the EAS, hence we no need to the config changes. updated config file.

PR #872, opened 2020-02-20
CON-1084-update GTM code

shomavg:


src/Employer/Employer.Web/Views/Dashboard/_NoVacanciesContent.cshtml
```diff
@@ -1,5 +1,8 @@
 ﻿@using Microsoft.Extensions.Options
 @inject IOptions<ExternalLinksConfiguration> _externalLinks
+@{ 
+    ViewBag.Vpv = "/page-no-vacancies";
```
cofaulco: the naming of this file suggests that it is a partial. In which case should it have a vpv?

shomavg: i will remove this bit of code.

PR #873, opened 2020-03-04
CON-1125 contextual widget

shomavg:


src/Employer/Employer.Web/Views/Shared/_Layout.cshtml
```diff
@@ -89,30 +89,31 @@
...
+            </a>
+        </div>
+        <div class="govuk-header__content">
+            <a href="@ExternalLinks.MaRoot" class="govuk-header__link govuk-header__link--service-name">
+                Manage apprenticeships
+            </a>
         </div>
-    </header>
+    </div>
+    @Html.Partial("_ZenDeskWidget")
```
Stephek: Don't put this in the header.

src/Employer/Employer.Web/Views/Shared/_Layout.cshtml
```diff
@@ -89,30 +89,31 @@
 
     @Html.Partial("_CookieBanner")
 
-    <header class="govuk-header " role="banner" data-module="header">
```
Stephek: Don't change the header.  It's being replaced with the global one.

src/Shared/Recruit.Shared.Web/Configuration/ZendeskConfiguration.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Esfa.Recruit.Shared.Web.Configuration
+{
+    public sealed class ZendeskConfiguration
+    {
+        public string ZenDeskSnippetKey { get; set; }
```
Stephek: Can these properties drop the ZenDesk prefix? They are in the ZendeskConfiguration class so it isn't needed.

src/Employer/Employer.Web/Startup.Configure.cs
```diff
@@ -43,17 +44,50 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, IOptions
...
                 app.UseHsts(hsts => hsts.MaxAge(365));
             }
-            
+
+            if (env.IsDevelopment())
+            {
+                IdentityModelEventSource.ShowPII = true;
+            }
             // Add Content Security Policy
+            app.Use(async (context, next) =>
```
Stephek: Can all the Zendesk config be put into it's own Use extension?  e.g. app.UseZenDesk ?

src/Employer/Employer.Web/Startup.Configure.cs
```diff
@@ -43,17 +44,50 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, IOptions
...
+            {
+                context.Response.Headers["Content-Security-Policy"] =
+                    "default-src 'self' 'unsafe-inline' https://*.zdassets.com https://*.zendesk.com wss://*.zendesk.com wss://*.zopim.com; " +
+                    "img-src 'self' *.google-analytics.com https://*.zdassets.com https://*.zendesk.com wss://*.zendesk.com wss://*.zopim.com; " +
+                    "script-src 'self' 'unsafe-inline' " +
+                    "*.googletagmanager.com *.postcodeanywhere.co.uk *.google-analytics.com *.googleapis.com https://*.zdassets.com https://*.zendesk.com wss://*.zendesk.com wss://*.zopim.com; " +
+                    "font-src 'self' data:;";
+                await next();
+            });
```
cofaulco: Do we need this middleware to add the header if we are already have the UseCsp middleware below setting all the same values?

azure/template.json
```diff
@@ -915,6 +927,18 @@
...
                 "name": "PostcodeAnywhere:Key",
                 "value": "[parameters('PostcodeAnywhereKey')]"
+              },
+			  {
+                "name": "Zendesk:SnippetKey",
+                "value": "[parameters('ZendeskSnippetKey')]"
+              },
+			  {
+                "name": "Zendesk:SectionId",
+                "value": "[parameters('ZendeskSectionId')]"
```
cofaulco, line 937: Might be best to name this ZendeskEmployerSectionId as they are different for employer and provider.

shomavg: 👍

PR #891, opened 2020-06-25
CON-1977 content and design changes to create an advert second half

ben1stone:
Updates content and design on the Pay, working hours and preview start page.

The updated VacacnyServices.Wage package was to change the content that is returned from the GetYearlyMinimumWageText method

### DIFF
--- a/src/Employer/Employer.Web/Views/Part1/Duration/Duration.cshtml
+++ b/src/Employer/Employer.Web/Views/Part1/Duration/Duration.cshtml
@@ -44,7 +44,7 @@
                     <span class="govuk-hint">For example, Monday to Friday 9am to 5pm, shifts, may work evenings and weekends.</span>
                     <span esfa-validation-message-for="WorkingWeekDescription" class="govuk-error-message"></span>
                     <textarea asp-for="WorkingWeekDescription" class="govuk-textarea character-count" rows="5" data-val-length-max="250"></textarea>
-                    <span class="govuk-body maxchar-text">You have </span>
+                    <span class="govuk-body">You have </span>
                     <span class="govuk-body maxchar-count">250</span>
                     <span class="govuk-body maxchar-text"> characters remaining</span>
                 </div>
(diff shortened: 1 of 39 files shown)


src/Employer/Employer.Web/Views/Part1/Duration/Duration.cshtml
```diff
@@ -32,7 +32,7 @@
                     <span esfa-validation-message-for="Duration" class="govuk-error-message"></span>
                     <input asp-for="Duration" class="govuk-input govuk-input--width-2" type="number" min="1" >
                     <select asp-for="DurationUnit" class="govuk-select duration-unit-select">
-                        <option value="@DurationUnit.Month">Month(s)</option>
+                        <option value="@DurationUnit.Month">Months</option>
                         <option value="@DurationUnit.Year">Year(s)</option>
```
shomavg, line 36: why not make the same change for Years too?

src/Shared/Recruit.Vacancies.Client/Recruit.Vacancies.Client.csproj
```diff
@@ -37,7 +37,7 @@
     <PackageReference Include="RestSharp" Version="106.6.10" />
     <PackageReference Include="SFA.DAS.Account.Api.Client" Version="1.6.1601" />
     <PackageReference Include="SFA.DAS.Http" Version="1.2.4" />
-    <PackageReference Include="SFA.DAS.VacancyServices.Wage" Version="1.0.8" />
+    <PackageReference Include="SFA.DAS.VacancyServices.Wage" Version="1.0.13" />
```
shomavg, line 40: why is the wage updated?

ben1stone: The updated VacacnyServices.Wage package was to change the content that is returned from the GetYearlyMinimumWageText method

src/Employer/Employer.Web/Views/Part1/Duration/Duration.cshtml
ben1stone: Apprenticeships are minimum 12 Months long so the optional (s) doesn't make sense but they could choose to write that as 1 Year.

src/Shared/Recruit.Vacancies.Client/Recruit.Vacancies.Client.csproj
shomavg: 👍

src/Employer/Employer.Web/Views/Part1/Duration/Duration.cshtml
shomavg: 👍

PR #892, opened 2020-06-26
CON-1988 : Content and design changes to Skills and personal qualitie…

VasanthaKasirajan3008:


src/Shared/Recruit.Vacancies.Client/Application/Validation/Fluent/FluentVacancyValidator.cs
```diff
@@ -410,7 +410,7 @@ private void ValidateSkills()
         {
             RuleFor(x => x.Skills)
                 .Must(s => s != null && s.Count > 0)
-                    .WithMessage("You must include a skill or quality")
+                    .WithMessage("Select the skills and personal qualities you�d like the applicant to have")
```
cofaulco, line 413: what is the special character in this string?

PR #906, opened 2020-07-23
CON-2073-change to advert.

shomavg:


src/Employer/Employer.Web/Configuration/Routing/RoutePaths.cs
```diff
@@ -5,7 +5,7 @@ public static class RoutePaths
         public const string Services = "services";
         public const string AccountLevelServices = "services/{employerAccountId:minlength(6)}";
         public const string AccountRoutePath = "accounts/{employerAccountId:minlength(6)}";
-        public const string VacanciesRoutePath = AccountRoutePath + "/vacancies";
+        public const string VacanciesRoutePath = AccountRoutePath + "/your-adverts";
```
cofaulco, line 8: I believe the story has a note on to say that url changes are not to be included in this work

PR #907, opened 2020-07-23
CON-2091-make vacancy as advert.

shomavg:


src/Employer/Employer.Web/Controllers/VacancyManageController.cs
```diff
@@ -26,7 +26,7 @@ public VacancyManageController(VacancyManageOrchestrator orchestrator, IHostingE
             _hostingEnvironment = hostingEnvironment;
         }
 
-        [HttpGet("manage", Name = RouteNames.VacancyManage_Get)]
+        [HttpGet("manage-advert", Name = RouteNames.VacancyManage_Get)]
```
cofaulco, line 29: I thought we weren't changing url's as part of this work? As it is going to throw off all the google analytics

PR #944, opened 2021-02-18
CON-2779 Restrict Viewer Users

Howesy77:


src/Provider/Provider.Web/Configuration/ServiceClaims.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Esfa.Recruit.Provider.Web.Configuration
+{
+    public enum ServiceClaim
+    {
+        DAA = 4,
```
cofaulco, line 5: might be worth increasing the numeric values like in reservations. Guards against having to change the values if a new Service claim added between `DAA` and `DAB for example

src/Provider/UnitTests/Provider.Web/Extensions/ServiceClaimExtensionsTests.cs
```diff
@@ -0,0 +1,145 @@
...
+            var result = claimsPrincipal.HasPermission(minimumServiceClaim);
+            result.Should().Be(expected);
+        }
+
+        [Theory]
+        [InlineData(ServiceClaim.DAA, false)]
+        [InlineData(ServiceClaim.DAB, true)]
+        [InlineData(ServiceClaim.DAC, true)]
+        [InlineData(ServiceClaim.DAV, true)]
+        public void ShouldReturnHasPermission_DAB(ServiceClaim minimumServiceClaim, bool expected)
```
cofaulco, line 74: Could these all be done in 1 test with another param in the inline data? minimumServiceClaim, actualServiceClaim, expected. Or something along these lines? Removes the need to know which claims to add/remove from the List<Claim>

cofaulco (commented):
A couple of minor comments

cofaulco (approved):

PR #951, opened 2021-04-15
CON-3392 Updated content for access denied

ChrisJWoodcock:


src/Provider/Provider.Web/Views/Shared/_Layout.cshtml
```diff
@@ -19,7 +19,11 @@
     {
         const string defaultTitle = "Recruit an Apprentice";
         if (ViewBag.Title != null)
-            return $"{defaultTitle} | {ViewBag.Title}";
+        { 
+            return !(ViewBag.HideDefaultTitle ?? false)
```
cofaulco, line 23: might want to run this one by BA / content, if this is how all of Recruits page titles are handled

cofaulco (commented):
1 comment

src/Provider/Provider.Web/Views/Shared/_Layout.cshtml
ChrisJWoodcock: I did run it by Annukka - originally it was going to keep the default title as Paul said it was hard to remove - but they did want it to match - so I made this change

ChrisJWoodcock: Probably the whole of recruit should be changed - which would be easy enough as it's the odd one out

cofaulco (approved):

PR #962, opened 2021-05-15
Con 3314 approve job advert

VasanthaKasirajan3008:


src/Employer/Employer.Web/Mappings/VacancyDisplayViewModelMapper.cs
```diff
@@ -42,7 +42,7 @@ public async Task MapFromVacancyAsync(DisplayVacancyViewModel vm, Vacancy vacanc
             vm.ApplicationInstructions = vacancy.ApplicationInstructions;
             vm.ApplicationUrl = vacancy.ApplicationUrl;
             vm.CanDelete = vacancy.CanDelete;
-            vm.CanSubmit = vacancy.CanSubmit;
+            vm.CanSubmit = vacancy.CanSubmit && vacancy.Status != VacancyStatus.Review;
```
cofaulco, line 45: should this check be moved to the vacancy domain object if it is a behaviour of all vacancies? Then the mapping to the view model would be straight `vm.CanSubmit = vacancy.CanSubmit`?

cofaulco (commented):
1 comment

src/Employer/Employer.Web/Controllers/VacancyPreviewController.cs
```diff
@@ -70,6 +71,59 @@ public async Task<IActionResult> Submit(SubmitEditModel m)
...
+
+        [HttpPost("approve-advert", Name = RouteNames.ApproveJobAdvert_Post)]
+        public async Task<IActionResult> ApproveJobAdvert(ApproveJobAdvertViewModel vm)
+        {
+            if (!ModelState.IsValid)
+            {
+                return View("ApproveJobAdvert");
+            }
+
+            if ((bool)vm.ApproveJobAdvert)
```
cofaulco, line 88: why the cast rather than just `vm.ApproveJobAdvert` or `vm.ApproveJobAdvert.GetValueOrDefault()`

cofaulco (commented):
additional comment

src/Employer/Employer.Web/Controllers/VacancyPreviewController.cs
VasanthaKasirajan3008: done

src/Employer/Employer.Web/Mappings/VacancyDisplayViewModelMapper.cs
VasanthaKasirajan3008: introduced this check to get the  correct link displayed on the page either Submit or Approve Job Advert... 
I got  viewmodel  mapper CanReview does the job. 
i have removed this check. nice one to get this sorted.

cofaulco (approved):

PR #990, opened 2021-09-06
CON-3892 Issue with Employer Name and Description in recruit

Howesy77:


ChrisJWoodcock (approved):
We don't know why this code was added originally and explicitly commented as when there is no employer description the vacancy cannot be submitted

PR #1191, opened 2022-08-22
FAI-556 application missing from vacancy

dashton82:
This removes the need for the QueryStore view that shows the applications. The logic for creating and maintaining has been maintained in the code base should it need to be changed back due to performance issues.


scottywakefield (approved):

dashton82:
This has been done in another PR

PR #1202, opened 2022-09-13
Add style to preserve white space

A-J-Wright:


jamesmking (commented):
This is a good solution, but its not good practice to use inline styles (`style="..."`), it would be preferable if we could create a CSS class in the local CSS file (app.css) in each project, something like:
`.app-preserve-line-breaks {
    white-space: pre-line;
}`

And then add that class where you need it, e.g. `<p class="govuk-body app-preserve-line-breaks">`

dashton82 (approved):

PR #1234, opened 2022-12-02
Qa service users fai 667

balaji-j:
### Story: The purpose of this story is to put behind a feature toggle the DfE sign in authentication, replacing the existing Pirean, SAML based authentication.
Link: https://skillsfundingagency.atlassian.net/browse/FAI-667

**Details**
1) Extended SFA.DAS.Recruit.QA Config to hold the UseDfeSignIn property.  "UseDfeSignIn": true.
2) Added DfeSignIn Base Url in the AllowDestinationUrl so that the redirect could be allowed by QA.Web application. 
3) Added logic/condition in ServiceCollectionExtension to check if the UseDfeSignIn property is set to true, so that DfESignin happens. 
4) Check the Authentication scheme based on UseDfeSignIn property and use the relevant sign-out Authentication Scheme.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -20,12 +20,12 @@ resources:
   - repository: das-platform-building-blocks
     type: github
     name: SkillsFundingAgency/das-platform-building-blocks
-    ref: refs/tags/0.4.61
+    ref: refs/tags/2.1.0
     endpoint: SkillsFundingAgency
   - repository: das-platform-automation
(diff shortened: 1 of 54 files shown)


balaji-j:
Updated the PR with DfeSignIn Nuget package included

https://www.nuget.org/packages/SFA.DAS.DfeSignIn.Auth

dashton82 (approved):

PR #1274, opened 2023-06-14
Ec 63 banner

DavidRimarDfE:


src/Shared/Recruit.Shared.Web/ViewModels/InfoMessages.cs
```diff
@@ -19,5 +19,13 @@ public class InfoMessages
         public const string AdvertCloned = "Advert succesfully cloned";
 
         public const string ApplicationReviewStatusHeader = "{0} application has been marked as {1}";
+
+        public const string SharedMultipleApplicationsBannerHeader = "Multiple applications shared with employer.";
+
+        public const string SharedMultipleApplicationsBannerMessage = "These applications have been sent to the employer's apprenticeship service account.";
+
+        public const string SharedSingleApplicationsBannerHeader = "{0}'s application shared with Employer";
```
chrisfoster76, line 27: I notice that Employer has a capital E but does not in other headers. Could be per spec tho.

chrisfoster76 (approved):

PR #1331, opened 2023-08-18
apply sorting on multiple applications table employer

DavidRimarDfE:


src/Employer/UnitTests/Employer.Web/Controllers/ApplicationReviewsControllerTests.cs
dashton82, line 1: Would be good if we could also get a test in to cover the flow of what happens if the sort column and sort order arent valid enum values

src/Employer/Employer.Web/Controllers/ApplicationReviewsController.cs
```diff
@@ -26,20 +27,26 @@ public ApplicationReviewsController(IApplicationReviewsOrchestrator orchestrator
...
+
+            var viewModel = await _orchestrator.GetApplicationReviewsToUnsuccessfulViewModelAsync(rm, outputSortColumn, outputSortOrder);
 
             return View(viewModel);
         }
 
         [FeatureGate(FeatureNames.MultipleApplicationsManagement)]
         [HttpPost("unsuccessful", Name = RouteNames.ApplicationReviewsToUnsuccessful_Post)]
-        public async Task<IActionResult> ApplicationReviewsToUnsuccessfulAsync(ApplicationReviewsToUnsuccessfulRouteModel rm)
+        public async Task<IActionResult> ApplicationReviewsToUnsuccessfulAsync(ApplicationReviewsToUnsuccessfulRouteModel rm, [FromQuery] string sortColumn, [FromQuery] string sortOrder)
```
dashton82, line 42: is there a reason why this isnt on the model being posted? for the sortColumn and sortOrder?

DavidRimarDfE: The reason is that if the user does not select any checkboxes, then the server-side sorting is "preserved", it basically goes to the db again to get the same applications sorted in the same order. Not sure if there is a way around it? Open to suggestions. Maybe the tech debt ticket EC-114 would be a solution, removing the modelstate validations?

src/Employer/UnitTests/Employer.Web/Controllers/ApplicationReviewsControllerTests.cs
DavidRimarDfE: added a test case in my last commit, it'll basically default to the Default enum value, amended the setup to match what it's gonna get

src/Employer/Employer.Web/Controllers/ApplicationReviewsController.cs
DavidRimarDfE: No reason. I moved it inside the posted model

dashton82 (approved):

PR #1377, opened 2023-10-26
FAI-971 Add get all live vacancies endpoint

SarahDuncan:
Add endpoint to get all vacancies:
-  that have a closing date that is in the future, 
- and sort these results by closing date (in descending order)
- with a default of 100 results per page,
- and a maximum of 1000 results returned (even if the paging numbers would expect more)

### DIFF
--- a/src/API/Recruit.Api/Queries/GetLiveVacanciesQueryHandler.cs
+++ b/src/API/Recruit.Api/Queries/GetLiveVacanciesQueryHandler.cs
@@ -1,14 +1,30 @@
-﻿using System;
+﻿using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using Esfa.Recruit.Vacancies.Client.Infrastructure.QueryStore;
 using MediatR;
 
 namespace SFA.DAS.Recruit.Api.Queries;
 
 public class GetLiveVacanciesQueryHandler : IRequestHandler<GetLiveVacanciesQuery, GetLiveVacanciesQueryResponse>
 {
+    private readonly IQueryStoreReader _queryStoreReader;
+
+    public GetLiveVacanciesQueryHandler(IQueryStoreReader queryStoreReader)
+    {
+        _queryStoreReader = queryStoreReader;
+    }
+
     public async Task<GetLiveVacanciesQueryResponse> Handle(GetLiveVacanciesQuery request, CancellationToken cancellationToken)
     {
-        throw new NotImplementedException();
+        var queryResult = await _queryStoreReader.GetAllLiveVacancies();
+
+        if (queryResult == null) { return new GetLiveVacanciesQueryResponse { ResultCode = Models.ResponseCode.NotFound }; }
+
+        //var numberResults = queryResult.ToList().Count();
+
+        return new GetLiveVacanciesQueryResponse { ResultCode = Models.ResponseCode.Success, Data = queryResult.ToList() };
+
+        // implement paging
     }
 }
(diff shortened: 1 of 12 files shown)


src/API/Recruit.Api/Queries/GetLiveVacanciesQueryHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+        _queryStoreReader = queryStoreReader;
+    }
+
+    public async Task<GetLiveVacanciesQueryResponse> Handle(GetLiveVacanciesQuery request, CancellationToken cancellationToken)
+    {
+        var liveVacanciesToFetch = request.PageSize * request.PageNumber;
+
+        var queryResult = liveVacanciesToFetch > 1000 ? await _queryStoreReader.GetAllLiveVacancies(1000) : await _queryStoreReader.GetAllLiveVacancies(liveVacanciesToFetch);
+
+        if (queryResult == null) { return new GetLiveVacanciesQueryResponse { ResultCode = ResponseCode.NotFound }; }
```
dashton82, line 26: I think for this we should still return success but an empty array instead

src/API/Recruit.Api/Queries/GetLiveVacanciesQueryHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        var totalLiveVacancies = queryResult.Count();
```
dashton82, line 28: This requires a separate call to get the total live vacancies, there should be the equivalent of count in the mongo client which you can put the same filter into

src/API/Recruit.Api/Queries/GetLiveVacanciesQueryHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+    private readonly IQueryStoreReader _queryStoreReader;
+
+    public GetLiveVacanciesQueryHandler(IQueryStoreReader queryStoreReader)
+    {
+        _queryStoreReader = queryStoreReader;
+    }
+
+    public async Task<GetLiveVacanciesQueryResponse> Handle(GetLiveVacanciesQuery request, CancellationToken cancellationToken)
+    {
+        var liveVacanciesToFetch = request.PageSize * request.PageNumber;
```
dashton82, line 22: I think this should be the Vacancies to skip, as in Ive requested page 5, and my page size is 1000, then i need to skip the first 5000 records and then get from 5001-6000 for my next page

src/API/Recruit.Api/Queries/GetLiveVacanciesQueryHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        var queryResult = liveVacanciesToFetch > 1000 ? await _queryStoreReader.GetAllLiveVacancies(1000) : await _queryStoreReader.GetAllLiveVacancies(liveVacanciesToFetch);
```
dashton82, line 24: I think this will fail when i have a page size of 1000 and request anything over page 1

src/Shared/Recruit.Vacancies.Client/Infrastructure/QueryStore/MongoQueryStore.cs
```diff
@@ -152,5 +152,22 @@ Task IQueryStore.UpsertAsync<T>(T item)
...
+            var builderFilter = Builders<LiveVacancy>.Filter;
+            var filter = builderFilter.Gt(identifier => identifier.ClosingDate, DateTime.UtcNow);
+
+            var builderSort = Builders<LiveVacancy>.Sort;
+            var sort = builderSort.Descending(identifier => identifier.ClosingDate);
+
+            var collection = GetCollection<LiveVacancy>();
+
+            var result = await RetryPolicy.Execute(_ =>
+                    collection.Find(filter).Sort(sort).Limit(numberToGet).Project<LiveVacancy>(GetProjection<LiveVacancy>()).ToListAsync(),
```
dashton82, line 167: You also need the `Skip()` here which will be the page size * page number

dashton82 (changes requested):
All looks good, but requires a few changes on the paging logic

dashton82 (approved):

PR #1387, opened 2023-12-18
Fai 975 sorting in faa using azure cognitive search

chrisfoster76:


chrisfoster76:
Closed as created with wrong name

PR #1393, opened 2024-01-03
Filter change to exclude closed vacancies from result

chrisfoster76:


dashton82 (approved):

dashton82:
already in as part of another ticket

PR #1447, opened 2024-04-30
FAI-1187 - Add submit application endpoint for recruit

dashton82:


dashton82:
Already included as part of another PR

PR #1546, opened 2025-01-06
FAI-2090 - Create the add more than one location from a list page   provider

killij:


src/Provider/Provider.Web/Services/VacancyLocationService.cs
```diff
@@ -0,0 +1,59 @@
...
+
+public class VacancyLocationService(IRecruitVacancyClient recruitVacancyClient, IEmployerVacancyClient employerVacancyClient): IVacancyLocationService
+{
+    public async Task<List<Address>> GetVacancyLocations(Vacancy vacancy)
+    {
+        ArgumentNullException.ThrowIfNull(vacancy);
+        var employerProfile = await recruitVacancyClient.GetEmployerProfileAsync(vacancy.EmployerAccountId, vacancy.AccountLegalEntityPublicHashedId);
+
+        var employerData = await employerVacancyClient.GetEditVacancyInfoAsync(employerProfile.EmployerAccountId);
+        var legalEntity = employerData.LegalEntities.First(l => l.AccountLegalEntityPublicHashedId == employerProfile.AccountLegalEntityPublicHashedId);
```
dashton82, line 28: i think i would be tempted to change this to FirstOrDefault - only to cater for scenarios where the AccountLegalEntity is removed from the account by the employer.

dashton82 (approved):

PR #1550, opened 2025-01-07
FAI-2168 Skip WithDrawn Status Vacancies in the Sharing Applications

balaji-j:
Story: https://skillsfundingagency.atlassian.net/browse/FAI-2168


killij (approved):
Just checking that unsuccessful candidates are excluded by default?

dashton82 (approved):

PR #1567, opened 2025-01-17
FAI-2236 correct vacancy count

dashton82:


src/Shared/Recruit.Vacancies.Client/Infrastructure/Services/VacancySummariesProvider/VacancySummaryAggQueryBuilder.cs
```diff
@@ -459,26 +459,61 @@ public class VacancySummaryAggQueryBuilder
...
+                    'preserveNullAndEmptyArrays': true
+                }
+            },    
+            {
+                '$project': {
+                    'vacancyGuid': '$_id',
+                    'vacancyReference': 1
+                }
+            },
+            {
```
killij, line 485: Is this essentially the same as the previous step in the pipeline?

killij (approved):

PR #1656, opened 2025-05-16
Fai 2512 foundation apprenticeship additional questions

LACrossley:


src/Provider/Provider.Web/Views/Part2/AdditionalQuestions/AdditionalQuestions.cshtml
```diff
@@ -14,22 +14,43 @@
...
+
         <h1 class="govuk-heading-xl">
-            Do you have any questions you would like to ask applicants? (optional)
+            Application questions on Find an apprenticeship
         </h1>
-        <p class="govuk-body">Applicants will answer these questions if they apply through <a href="@Model.FindAnApprenticeshipUrl" class="govuk-body govuk-link" rel="noopener" target="_blank">Find an apprenticeship (opens in a new tab)</a>.</p>
-        <p class="govuk-body">For example, 'How will you travel to this apprenticeship?'</p>
+        <p class="govuk-body">Find an apprenticeship includes set questions that applicants will need to answer. You can add extra questions if you need to ask more.</p>
+
+        @if (Model.ApprenticeshipType == ApprenticeshipTypes.Foundation)
```
dashton82, line 24: Same as above, we can simplify this and reduce the duplication

src/Shared/Recruit.Shared.Web/Views/Shared/_ValidationSummaryPartial.cshtml
```diff
@@ -11,9 +11,13 @@
...
                                                        .OrderBy(x => Model.OrderedFieldNames.IndexOf(x.Key)))
             {
-                var errorVm = new ErrorListItemViewModel(ViewData.TemplateInfo.GetFullHtmlFieldName(stateItem.Key) ?? stateItem.Key,
-                    stateItem.Value.Errors.FirstOrDefault()?.ErrorMessage);
-                <partial name="@RecruitPartialNames.ErrorListItem" model="@errorVm" />
+                var fieldName = ViewData.TemplateInfo.GetFullHtmlFieldName(stateItem.Key) ?? stateItem.Key;
+
+                foreach (var error in stateItem.Value.Errors)
+                {
+                    var errorVm = new ErrorListItemViewModel(fieldName, error.ErrorMessage);
```
dashton82, line 18: This _looks_ like its doing the wrong thing and we will end up with duplicated error messages

src/Shared/Recruit.Vacancies.Client/Application/Validation/Fluent/FluentVacancyValidator.cs
```diff
@@ -605,38 +605,56 @@ private void ValidateAdditionalQuestions()
             When(x => !string.IsNullOrEmpty(x.AdditionalQuestion1), () =>
             {
                 RuleFor(x => x.AdditionalQuestion1)
+                    .Cascade(CascadeMode.Continue)
```
dashton82, line 608: Links to the above with the new show-all attribute on the validation tag helper, but where is the requirement for this?

dashton82 (changes requested):

src/Provider/Provider.Web/Orchestrators/Part2/AdditionalQuestionsOrchestrator.cs
```diff
@@ -43,10 +44,13 @@ public async Task<AdditionalQuestionsViewModel> GetViewModel(VacancyRouteModel r
...
         {
             VacancyId = vacancy.Id,
+            VacancyTitle = vacancy.Title,
             Ukprn = routeModel.Ukprn,
             AdditionalQuestion1 = vacancy.AdditionalQuestion1,
             AdditionalQuestion2 = vacancy.AdditionalQuestion2,
-            FindAnApprenticeshipUrl = _options.Value.FindAnApprenticeshipUrl
+            FindAnApprenticeshipUrl = _options.Value.FindAnApprenticeshipUrl,
+            ApprenticeshipType = (ApprenticeshipTypes)vacancy.ApprenticeshipType,
+            QuestionCount = (ApprenticeshipTypes)vacancy.ApprenticeshipType == ApprenticeshipTypes.Foundation ? 3 : 4
```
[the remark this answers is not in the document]
killij: There is an extension GetValueOrDefault() for ApprenticeshipTypes to handle this scenario.

src/Shared/Recruit.Shared.Web/Views/Shared/_ValidationSummaryPartial.cshtml
LACrossley: Each applicable error may only be triggered once per field either way. So currently, the same message may appear twice in the banner if it applies to two different fields, and the links take the user to each field. That behaviour doesn't change with this.

It's true the story doesn't directly request this change. However, without it, the validation summary box only shows one of the applicable errors per field, which feels like unintentionally bad UX design. Same without "show-all" for messages shown on the field. So previously, the user had to click submit each time to be told "over 250 characters", click, "no question mark", click, "and now you've used bad language" even though all three applied to the same field at the same time.

With this code change, the user can now click submit once to be told "question 3 is too long, it has no '?', and you've used a naughty word". So they can fix all three before hitting submit once more successfully.

I'll leave this in for now and I've just asked Chris about it. I appreciate this would affect the whole of recruit so I'll revert the change if need be.

src/Provider/Provider.Web/Views/Part2/AdditionalQuestions/AdditionalQuestions.cshtml
LACrossley: Amended in latest commit.

src/Provider/Provider.Web/Orchestrators/Part2/AdditionalQuestionsOrchestrator.cs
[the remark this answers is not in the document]
LACrossley: Amended in latest commit.

src/Shared/Recruit.Vacancies.Client/Application/Validation/Fluent/FluentVacancyValidator.cs
LACrossley: In either case, CascadeMode.Continue doesn't appear to be as necessary as I thought. It still shows all errors without setting this. So I've removed from the latest commit.

dashton82:
Closing as its been recreated

PR #1719, opened 2025-08-01
FAI-2424 Add pagination to the Vacancy manage page

balaji-j:
✨ Add pagination to vacancy management functionality

- Updated `VacancyManageController` to support pagination with `page` parameter.
- Enhanced `GetManageVacancyViewModel` in `VacancyManageOrchestrator` for pagination.
- Modified `VacancyApplicationsViewModel` to include `Pager` property.
- Updated Razor views to include pagination controls for applications.
- Adjusted `application.js` to reset `page` parameter on location filter change.

Changes made by Balaji Jambulingam

### DIFF
--- a/src/Employer/Employer.Web/Orchestrators/VacancyManageOrchestrator.cs
+++ b/src/Employer/Employer.Web/Orchestrators/VacancyManageOrchestrator.cs
@@ -9,11 +9,13 @@ using Esfa.Recruit.Vacancies.Client.Application.Validation;
 using Microsoft.Extensions.Logging;
 using System.Collections.Generic;
 using System.Linq;
+using Esfa.Recruit.Employer.Web.Configuration.Routing;
 using Esfa.Recruit.Vacancies.Client.Infrastructure.QueryStore.Projections.VacancyApplications;
 using Esfa.Recruit.Vacancies.Client.Domain.Extensions;
 using Esfa.Recruit.Shared.Web.Orchestrators;
 using Esfa.Recruit.Shared.Web.Extensions;
 using Esfa.Recruit.Shared.Web.Helpers;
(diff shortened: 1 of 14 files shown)


src/Employer/Employer.Web/Orchestrators/VacancyManageOrchestrator.cs
```diff
@@ -69,21 +77,41 @@ public async Task<ManageVacancyViewModel> GetManageVacancyViewModel(Vacancy vaca
                 viewModel.WithdrawnDate = vacancy.ClosedDate?.AsGdsDate();
             }
 
-            var vacancyApplicationsTask = await _client.GetVacancyApplicationsSortedAsync(vacancy.VacancyReference.Value, sortColumn, sortOrder, vacancySharedByProvider);
-            var applications = vacancyApplicationsTask ?? new List<VacancyApplication>();
+            var vacancyApplications = await _client.GetVacancyApplicationsSortedAsync(vacancy.VacancyReference.Value, sortColumn, sortOrder, vacancySharedByProvider);
```
killij, line 80: Shouldn't we be getting the paging from the inner service? The problem otherwise is that we still have the performance hit of returning all rows?

balaji-j: that was the orginal design to pull but the we another layer of filter (Filter by location), as the location details are getting merged in the APIM layer from Candidate, we wont able to do the pagination total count

balaji-j: https://github.com/SkillsFundingAgency/das-apim-endpoints/pull/2307
https://github.com/SkillsFundingAgency/das-recruit-api/pull/60

killij (approved):

PR #1763, opened 2025-10-06
FAI-2797 Remove Application review functionality

balaji-j:


src/Shared/Recruit.Vacancies.Client/Infrastructure/Client/VacancyClient.cs
```diff
@@ -451,7 +451,8 @@ public Task<IEnumerable<LegalEntity>> GetEmployerLegalEntitiesAsync(string emplo
 
         public Task CreateApplicationReviewAsync(Domain.Entities.Application application)
         {
-            return messaging.SendCommandAsync(new CreateApplicationReviewCommand { Application = application });
+            //return messaging.SendCommandAsync(new CreateApplicationReviewCommand { Application = application });
```
killij, line 454: Does this comment need removing?

killij: And if so does the method need removing?

killij (changes requested):

killij (approved):

